PR #27, opened 2026-08-18
feat: Facebook and Business translation adapter (clean)

9529360-cpu:
Clean replacement for stale PR #17, rebased onto the current production/account-control-plane master.

Final PR diff intentionally contains only the 7 Facebook feature files:
- `src/main.cjs`
- `test/facebook-translation-contract.cjs`
- `ui/app.js`
- `ui/facebook-translation-adapter.js`
- `ui/index.html`
- `ui/style.css`
- `ui/translation-core.js`

It excludes the old temporary Windows smoke workflow, branch-write permissions, smoke trigger file, subscription/login debugging code, and updater/window debugging files. Those general account/window fixes were separated and merged independently.

Security boundaries retained: Facebook/Messenger URL allowlist, owned WebView registration, focused native input validation, no `innerHTML` translation writes, no chat body logs, DOM reuse fingerprinting, and translated-send interception.

Encrypted-chat restore compatibility: when Facebook/Messenger presents a blocking PIN / one-time-code / secure-storage restore dialog, the translation adapter now stays passive: it does not scan message DOM, insert translation nodes, or intercept sends. The guard checks only visible modal geometry and does not read PINs, verification codes, or chat text. Translation resumes after the modal is removed.

Final CI passes on the cleaned branch. A Windows restore-guard smoke artifact was built successfully in run 32158281529 without repository, Cloudflare, Geek-account, or Facebook credentials; all temporary patch/build workflow files were removed afterward.

Keep draft until a real local Facebook + Business Inbox smoke test succeeds. No Facebook credentials are requested or stored by CI.


9529360-cpu:
Pausing this Facebook / Business Inbox work after repeated physical-device testing of Messenger E2EE restore. The current branch is not being merged: PIN / one-time-code restore remains unreliable in the embedded Electron environment despite isolated tests across preload/bridge removal, UA/client hints, popup/opener behavior, proxy mode, sandbox mode, and runtime comparisons. Keep the branch as reference for future investigation, but return development focus to the core Geek application. No credentials, PINs, OTPs, cookies, or test user-data are included in this PR.

Issue #132, opened 2026-08-20
[P1] LINE translation fails on the known-good 1.2.12 startup path


9529360-cpu:
## Current verified symptom

Production updater has been rolled back to **Geek 1.2.12** after 1.2.13 caused a real-client LINE white screen. On 1.2.12, authenticated LINE renders normally, can log in and can send ordinary messages, but Geek translation does not run.

## Correct investigation boundary

The previous proposed fix — synthesizing a partial page-visible `window.$electron = { send2Host }` from the active LINE preload — is disproven by real-client behavior and must not be reintroduced. It changed LINE bootstrap semantics and produced the 1.2.13 white-screen regression.

The 1.2.12 translation adapter already has a deliberate fallback transport:

- when a compatible pre-existing `$electron.send2Host` is available, it may use it;
- otherwise it emits `__GEEK_TRANSLATION_REQUEST__:<id>:<bridgeToken>` through the page console;
- `ui/app.js` registers the WebView `console-message` listener and routes that marker through the same `authorizeWebviewBridge(...)` and `processTranslationRequest(...)` host path.

Therefore an empty `resources/s3loYR.js` is **not by itself evidence of the translation root cause**. The investigation must locate the exact 1.2.12 hop where the request is lost: adapter installation/DOM detection -> console marker -> WebView console event -> bridge authorization/payload retrieval -> translation call -> guest resolution/rendering.

## Constraints

- preserve the 1.2.12-compatible LINE startup/preload behavior;
- no synthetic `$electron` global;
- no `contextIsolation`, sandbox, nodeIntegration, webSecurity, LINE extension, authentication/token/HMAC, Worker/provider/quota or package/release version changes unless evidence later proves a distinct root cause;
- no logging chat plaintext, translated text, bridge tokens, LINE auth headers, cookies or real user data;
- keep Draft PR #131 separate.

## Acceptance

A fix must identify and contract-test the specific failed translation hop without altering LINE startup behavior. Real authenticated-client verification must confirm incoming translation and translated-send while normal LINE login/send remain intact before any future client release.

9529360-cpu:
[...]

PR #145, opened 2026-08-20
fix: read LINE messageBoxId from actual hash route

9529360-cpu:
Fixes the first confirmed root cause in #132.

Root cause:
- bundled LINE 3.5.1 uses hash-history message routes shaped as `/:routeSegment/:messageBoxId`;
- the current adapter hard-codes `/chats/<id>`;
- on the real route `chatId()` returns an empty string, so incoming-message processing and translated-send handling exit before issuing any translation request.

Change:
- parse the second segment from the exact two-segment LINE hash route and decode it as the messageBox/chat id;
- add a focused contract tied to the bundled LINE route definition and forbid reintroducing the `/chats/` hard-code.

Scope intentionally does not change LINE preload, `$electron`, context isolation, sandboxing, authorization, translation provider/quota logic, package version, or release metadata. Diagnostic PR #144 was closed without merge after proving the bridge request/authorization/pending/resolve round trip works on the actual bundled LINE extension.

No client release is part of this PR.


9529360-cpu:
Superseded by release PR #151. The same narrow LINE hash-route parser fix and focused contract are now on `master` as part of the 1.2.14 stable-baseline release tree. Closing this draft without merge to avoid duplicating the change.

Issue #148, opened 2026-08-20
[P1] LINE image group-send has no LINE-specific attachment transport


9529360-cpu:
## Verified gap

On the known-good Geek 1.2.12 baseline, LINE has a dedicated text group-send adapter, but attachment sending falls through the generic non-WA path that was written around Telegram:

- host dispatches a real file drop and checks for Telegram-style `.modal-dialog` / `.modal-container` primary button state;
- renderer waits for a Telegram `Send 1 Files`-style modal and then invokes the platform adapter's `send` step;
- LINE's adapter `send` step only operates the ordinary `textarea-ex` composer and presses Enter; it has no attachment preview/upload/confirmation handling.

WhatsApp uses its separate WPP media path. Historical #98 confirms WA/TG image group-send worked on real Windows after the CDP guest-path fix in 1.2.10, so this issue must not regress or replace those paths.

## Investigation/fix boundary

Identify the actual bundled LINE 3.5.1 attachment input/upload/confirmation path and implement a LINE-specific attachment transport instead of reusing Telegram modal assumptions.

Constraints:
- preserve the 1.2.12-compatible LINE startup/preload behavior;
- do not change LINE contextIsolation/sandbox/nodeIntegration/webSecurity;
- preserve opaque file-token boundary; renderer must not regain real file paths or Base64 payloads;
- do not log filenames if avoidable, file contents, chat text, chat IDs, auth headers, tokens, cookies, or real user data;
- do not modify WA/TG media transports except for shared code strictly required by the proven root cause;
- no package/release marker change and no production release in this issue.

## Acceptance

1. Static/runtime evidence identifies LINE 3.5.1's actual attachment control/flow rather than guessed selectors.
2. LINE image group-send can attach and confirm a selected image on the existing 1.2.12-compatible startup path.
3. Failure is fail-closed: a failed attachment injection cannot fall through and send only the caption/text while reporting success.
4. Existing WA/TG attachment contracts remain unchanged/passing.
5. A local validation installer is tested with an authenticated LINE account before any future release.

9529360-cpu:
Confirmed root cause on exact Geek 1.2.12 baseline (`bbebd6c47dd88d8e61b241019e24fdf3ce768bef`) using bundled LINE 3.5.1 evidence:

1. Generic non-WA drop positioning only targets ProseMirror/contenteditable; LINE uses its own visible `data-is-dropzone=true` chat/file-modal dropzones and `textarea-ex` composer.
[...]

PR #149, opened 2026-08-20
test: inspect LINE 3.5.1 attachment transport

9529360-cpu:
Diagnostic-only PR for #148. It is intentionally based on the immutable Geek 1.2.12 release commit and adds only a temporary static probe/workflow. The probe reads the bundled LINE 3.5.1 `main.js` and source map and reports attachment-related source modules/identifiers so the product fix can use the extension's actual file/upload flow rather than guessed Telegram selectors.

No product runtime, account data, updater metadata, release marker, production service, security posture, or user data is changed. Do not merge this diagnostic PR; replace/close it once the exact LINE attachment flow is established.


9529360-cpu:
The diagnostic objective is complete: the real LINE 3.5.1 pasted-image/native editor submission path was established and authenticated V8 validation succeeded. The accepted implementation is now on `master` through release PR #151. Closing this diagnostic-only PR without merge as intended.

PR #309, opened 2026-09-10
fix: allow WhatsApp persistent storage in account session

9529360-cpu:
## Summary

Create a narrowly scoped validation candidate for the WhatsApp cold-start loading regression seen after v1.2.19.

The v1.2.19 account Session permission boundary changed Chromium permission handling from Electron defaults to a fail-closed allowlist. `persistent-storage` was explicitly denied for every account type. This PR grants only that permission to WhatsApp account Sessions, and only when the requesting origin already passes the existing WhatsApp navigation policy (`http://127.0.0.1:1843` or the allowed WhatsApp HTTPS origins).

## Why this candidate

- v1.2.18 did not bind the new Session permission boundary; v1.2.19 does.
- The reported failing client reaches `did-finish-load` and remains before native WhatsApp reaches its terminal chat/auth state, which is consistent with a post-document bootstrap/session boundary.
- Electron forwards Chromium permission checks through `setPermissionCheckHandler` / `setPermissionRequestHandler`; the current fail-closed matrix therefore changes runtime behavior for permissions that were not explicitly allowed.
- This is a single-variable, reversible candidate. It does not clear storage, recreate partitions, require QR re-login, disable security, or globally allow permissions.

## Security scope

Still denied:
- persistent storage for Telegram, LINE, and arbitrary Website accounts
- persistent storage outside the existing WhatsApp origin policy
- clipboard, display capture, devices, filesystem, geolocation, and every other unlisted permission

## Tests

`test/session-permission-boundary-contract.cjs` now covers request/check behavior for WhatsApp persistent storage and negative cross-origin/cross-platform cases.

## Validation boundary

This PR is **not yet proof of root cause**. Before calling the field regression fixed, validate on the existing affected Windows profile/session without clearing site data:

1. baseline v1.2.19 reproduces loading stall;
2. this exact-head validation build reaches native WhatsApp `chat-ready` or `auth-required`;
3. restoring the deny behavior reproduces the stall (A/B/A).

Do not change `.github/release-client-version` or publish a formal release as part of this PR.


9529360-cpu:
Real-client validation result (2026-09-10): the exact-head Windows validation build for `713f6b78bae737629f19ce4d61abffd7e3836d41` was installed over the existing affected profile/session without clearing WhatsApp data or requiring QR re-login, and the user reports WhatsApp now loads normally.

Evidence status:
- baseline formal v1.2.19: affected environment previously reproduced the Loading stall;
- candidate exact head: normal on the same existing session;
- CI: test + Electron smoke/E2E passed;
- isolated Windows validation build: passed and artifact produced;
- deny-restoration A/B/A leg was not performed, so this is strong field evidence for the repair but not a mechanistic proof that `persistent-storage` alone is the root cause.

Proceeding with the narrow fix; no formal release/version marker change in this PR.

PR #311, opened 2026-09-10
fix: update WA-JS cold-load loader to 4.5.0

9529360-cpu:
Validation candidate for the WhatsApp new-partition cold-start stall.

Evidence so far:
- formal Geek 1.2.20: existing WhatsApp sessions may load, but newly created WhatsApp partitions can remain on the native splash/loading screen;
- a today-rebuilt isolated client from the exact v1.2.18 runtime source (`ed21bca77290062c38e3a0b77be8c0f4c9ab216c`) also reproduces the new-partition stall, so the v1.2.19 account Session permission boundary is not sufficient to explain the failure;
- the fixed `resources/wa/index.html` still depends on live `static.whatsapp.net` assets, so an unchanged historical snapshot can encounter a different current cold-load/module graph;
- upstream WA-JS 4.5.0 specifically contains a cold-load module-graph timing fix for readiness getting stuck on current WhatsApp Web.

This PR is intentionally a single runtime variable: `@wppconnect/wa-js` 4.3.0 -> 4.5.0, with package version and `.github/release-client-version` remaining 1.2.20. No WebView security setting, session cleanup, account data, WA snapshot, or business logic is changed.

Do not merge based on CI alone. Required real-client gate: isolated exact-head Windows validation build, create a brand-new WhatsApp account/partition, and confirm whether the QR/auth-required screen is reached instead of stalling. If it still stalls, this candidate is refuted and should not be merged.


9529360-cpu:
Governance cleanup 2026-09-14: closing this historical validation candidate as refuted/superseded evidence, not as an unreviewed product merge. This PR tested WA-JS 4.5.0 as the single variable. The later descendant line (#315) added the WPP-settle ordering on top of that baseline, and #316 explicitly records real-client evidence that #315 still stalled on the native WhatsApp splash. Therefore the weaker #311 hypothesis is no longer a live merge candidate. Keep this PR/branch only for archaeology; any current WhatsApp cold-start fix must be re-established against live 1.2.24 source and current real-client evidence.

PR #314, opened 2026-09-10
debug: trace WhatsApp cold-load readiness in validation client

9529360-cpu:
Validation-only diagnostic candidate based on PR #311 (`@wppconnect/wa-js` 4.5.0).

Purpose: the real Windows client still showed the native WhatsApp splash/loading screen after the prior persistent-storage hypothesis. This candidate does not change WhatsApp behavior, permissions, security settings, account data, storage, or release/version markers. It adds a validation-profile-only trace on the real WhatsApp account WebContents.

The trace records only origin, document generation, error category, booleans and bounded counts for: document readiness, Meta loader globals/module count, WPP injected/ready/full-ready, conn main-ready, WAPLUS chat API presence, StorageManager persisted state, Service Worker support/control/registration count, IndexedDB support and localStorage item count. Console text is classified then discarded; no message body, cookie, token, login identifier, phone/email, or raw console text is logged.

Delayed probes are generation-guarded so navigation/reload invalidates stale results. Non-WhatsApp partitions and out-of-policy origins fail closed.

Prior art: WA-JS v4.5.0 / PR #3499 documents the cold-load failure where Meta modules register progressively, one-shot `registerStreamEvent` can run before `Stream` exists, and `isReady`/`isFullReady` then remain false. Electron's current WebContents navigation event uses a details object, so the trace accepts both current and deprecated event signatures.

Required gate before handing to user: exact-head `npm test`, Electron E2E, whitespace, then isolated Windows validation installer. Do not merge this diagnostic PR as a product fix.


9529360-cpu:
Governance cleanup 2026-09-14: this PR explicitly describes itself as validation-only diagnostics and says not to merge it as a product fix. The diagnostic lineage is now historical and the repository no longer uses this branch as current status. Closing without merge; retain the discussion/commits only as archaeology.

PR #315, opened 2026-09-10
fix: wait for WPP settle before WAPLUS injection

9529360-cpu:
## Root cause

Real-client startup diagnostics showed the WhatsApp guest could have `WAPLUS_WPP` and its chat API present while official WA-JS 4.5.0 still reported `WPP.isInjected === false`, followed by `MODULE_NOT_FOUND` / `UNDEFINED_ON` errors. `resources/waplus-wpp.js` embeds an older WA-JS 4.3.0 loader whose one-shot `loader.injected` callbacks can run before WhatsApp's Meta module graph is populated.

Upstream WA-JS 4.5.0 fixes this exact cold-load failure by waiting for the Meta module graph to settle before firing `loader.injected` (wppconnect-team/wa-js#3499).

## Change

Only change the existing `injectWppWithRetry()` owner in `src/main.cjs`:

1. inject official WA-JS 4.5.0;
2. if already injected, continue immediately;
3. otherwise wait on official `WPP.loader.onInjected()`;
4. only then inject `resources/waplus-wpp.js`.

No permission, storage, cache, service-worker, UI, account-data, version, or release changes.

## Validation so far

- one-shot patch guard: passed
- `node --check src/main.cjs`: passed
- `git diff --check`: passed
- `npm test`: passed
- final tree vs WA-JS 4.5.0 base: only `src/main.cjs`, +11/-0

## Real-client gate

Do not merge based on CI alone. Build the exact PR head as an isolated validation client and reproduce the previously stuck WhatsApp cold-load/account flow. Success requires reaching QR/auth/chat without the previous native WhatsApp splash stall and without the early `MODULE_NOT_FOUND` / `UNDEFINED_ON` pattern.


9529360-cpu:
Governance cleanup 2026-09-14: closing this historical product candidate as refuted by its own downstream real-client evidence. PR #316 was created after the exact #315 candidate still stalled on the native WhatsApp splash, so #315 did not satisfy its required real-client gate. Keep it for archaeology only; do not rebase/merge it into current master. Any current cold-start fix must be re-proven against live source and current real-client evidence.

PR #390, opened 2026-09-13
fix: keep broadcast stop transitions transactional

9529360-cpu:
## Root cause

Executing Broadcast Jobs committed `state=stopping` and `stopRequested=true` before awaiting the asynchronous stop control handler. If that handler rejected, `invoke(..., 'stop')` failed but the Job remained stuck in `stopping`, continued occupying the account execution slot, and the task-bar caller swallowed the error.

A naive handler-first fix would create a second race: two concurrent stop requests could both run the stop side effect before either request committed `stopping`.

## Change

- await the executing Job's stop control before committing `stopping`;
- if the control rejects, leave the prior running/paused/starting state and execution slot untouched so the stop can be retried safely;
- re-read the Job after the asynchronous control in case it already reached a terminal/stopping state;
- coalesce concurrent stop requests per Job so the stop handler runs at most once while a request is in flight;
- preserve pending scheduled/queued cancellation semantics and handler-less stop behavior.

## Regression coverage

New focused contract proves:

- rejected stop control does not commit `stopping` or `stopRequested`;
- the same-account execution slot remains occupied after a failed stop, preventing a second Job from racing the still-live executor;
- a later retry can successfully enter `stopping`;
- concurrent stop requests share one in-flight stop signal and do not duplicate handler side effects.

Focused local validation passed:

- `node --check ui/broadcast-job-manager.js`
- `node --check test/broadcast-stop-transaction-contract.cjs`
- `node test/broadcast-stop-transaction-contract.cjs` -> `BROADCAST_STOP_TRANSACTION_CONTRACT_OK`

The GitHub branch blobs match the locally tested blobs exactly (`1956cc0f...` and `f312689b...`).

## Collision boundary

Changed files are exactly intended to be:

- `ui/broadcast-job-manager.js`
- `test/broadcast-stop-transaction-contract.cjs`

The branch was created from live master `55b00b57fbe51bfd36894cf6e5ccf00e58b909a4`. Master later advanced to `69b724ce50fa89bb370d869061ac1540d5b4b2ee`, and `ui/broadcast-job-manager.js` remains unchanged there (same base blob `1ea483e0...`). No version/release marker, transport, attachment, WebView, external-CDP, config migration, translation, Worker, or package changes.

Require repository CI before merge.


9529360-cpu:
CI follow-up: the first exact-head Cloudflare validation failed because master advanced after branch creation and added `scripts/geek-website-entry.js`, while the older branch head did not yet contain that file. Repository tests themselves passed (170/170, including `BROADCAST_STOP_TRANSACTION_CONTRACT_OK`). I merged current master `69b724ce50fa89bb370d869061ac1540d5b4b2ee` into the branch without changing this PR's two-file diff. New exact head: `bdc32a6e8545115b49ed7e6863ee28ed0e92f95e`; fresh test / Cloudflare validation / Electron E2E runs are now triggered.

9529360-cpu:
Closing as already integrated by later master history. Both owned final blobs are already byte-identical on current `master`:

- `ui/broadcast-job-manager.js` = `1956cc0f6a04afe60d569543e95e42fac97da9d0`
- `test/broadcast-stop-transaction-contract.cjs` = `f312689b845841cd3f080ac2d3e1a933ee877991`

The transactional stop semantics and regression coverage are already present on master; merging this stale ancestry would add no product content.

Issue #395, opened 2026-09-13
[P1] Serialize desktop subscription state mutations to prevent concurrent temp-file races


9529360-cpu:
## Root cause

The desktop subscription store has one authoritative `subscription.json`, but its mutations are not serialized. `save()` and `clear()` can be invoked by independent async paths: after login the main process performs a background `store.refresh()`, while trusted renderer IPC can independently invoke refresh/get-quota/logout/login. Every write uses the same fixed `subscription.json.tmp` path.

Two mutations can therefore build from the same prior cache and concurrently write/rename the same temp file. Failure modes include:

- one operation renames the other operation's temp contents;
- the second rename sees ENOENT/EPERM/EBUSY;
- distinct patches overwrite each other because both candidates were derived before the first commit;
- logout/clear can race a save and leave disk/memory in an invocation-order-dependent state.

The mature `write-file-atomic` implementation explicitly queues concurrent writes to the same filename and serializes them in call order. Geek also needs to serialize the higher-level state transition, not only the final rename, so later patches are derived from the previously committed cache.

## Scope

- keep the existing state file, safeStorage encryption, temp-write + rename mechanism and API semantics;
- add one per-store mutation queue;
- run the full `save()` transition (load current -> build next -> encrypt/write -> commit cache) inside that queue;
- run `clear()` through the same queue so logout/account-switch cannot race a save;
- a failed mutation must reject but release the queue so later mutations can continue;
- reads remain non-blocking and observe the last committed cache while a mutation is in flight;
- no cross-process locking, new dependency, Worker/database/UI/main/WA/LINE/account-context/version/release change.

## Validation

Add an executable concurrency contract that drives public store methods and proves:
1. two concurrent state-writing operations never overlap the temp-file rename/write critical section;
2. distinct patches from concurrent operations are composed in queue order rather than losing the first committed patch;
3. a failed queued write rejects only that mutation and a later mutation still runs successfully;
4. concurrent clear/logout and save obey invocation order and never resurrect a cleared token from a stale candidate;
5. exact-head full repository, Cloudflare validation and Electron gates pass before merge.

Issue #387, opened 2026-09-13
[P1] Make translation quota reservation recoverable after unknown D1 outcomes


9529360-cpu:
## Root cause

The translation gateway already uses `X-Request-ID` as a durable idempotency key, but the initial source-character reservation is not recoverable when the D1 call outcome is unknown.

`reserveUsage()` currently runs a D1 batch that inserts `translation_usage(status='reserved')` and deducts source quota. Any thrown error is then collapsed to `duplicate_request`. If the batch actually committed and the caller only lost the response, the Worker returns 409 before contacting a provider, while the user's source quota remains deducted and the durable usage row remains `reserved` forever. A retry with the same request ID is also blocked by that row.

The same function also creates the usage row before discovering a zero-row quota update, then deletes the row in a separate statement. That cleanup is not tied to ownership of this reservation attempt.

Cloudflare's current D1 guidance says write retries must be business-idempotent, and D1 batches are transactions. Mature idempotency guidance likewise uses stable request IDs plus conditional writes/transaction state so an ambiguous write can be reconciled from durable state rather than guessed from an exception.

## Design

Use an attempt-owned reservation marker without a schema migration:

- store a server-random `reserved:<owner-id>` value in the existing `translation_usage.status` text field while a provider request owns the reservation;
- insert the usage row only when the authenticated user is active and has enough quota;
- deduct source quota only when the exact request/user/chars/owner marker exists in that same D1 batch;
- if the batch throws, read the durable request row: only an exact owner-marker match proves this attempt committed and may continue; a different row is a real duplicate; no row means the write did not establish a reservation and the original error must not be mislabeled as a duplicate;
- make finish/refund accept the exact owner marker so one concurrent request can never finish/refund another request's reservation;
- transition successful rows to the existing terminal `complete` state. No new table/column is required.

## Scope

- preserve source+target charging semantics, JWT/rate limits/provider selection/output checks and request-ID format;
- preserve 409 for real duplicate request IDs and 402 for genuine quota exhaustion;
- genuine pre-commit D1/storage errors should fail as translation failure rather than pretending to be duplicates;
- do not add a D1 migration or change Subscription/website/desktop/payment/version/release code.

## Validation

Extend the executable translation usage contract against the real Worker + SQLite-backed transactional D1 adapter to prove:

1. normal success still charges source+target and ends `complete`;
2. provider failure refunds only its own reservation and removes its own row;
[...]

PR #401, opened 2026-09-13
fix: cancel translation work when account is deleted

### LINKED ISSUE #397: [P1] Cancel account-owned translation work when the account is deleted
## Root cause

`src/translation-runtime.cjs::deleteAccount(partition)` currently calls `clearPartitionRuntimeState()`, which marks the partition deleted and removes cache/inflight/latest-request indexes. That prevents a successful late response from being cached, but it does not cancel the actual account-owned remote work.

The runtime has a global `remoteQueue` plus active `fetchImpl(..., { signal })` requests. Queue entries have no owner metadata, and the active AbortControllers are not tracked. As a result, after account A is deleted:

- A requests that were already queued can still start later and call the translation gateway;
- A requests already running continue until normal completion/timeout;
- clearing `state.inflight` only removes the lookup handle; it does not stop the Promise or network I/O;
- B account work must remain unaffected.

This violates the runtime owner's account lifecycle boundary. A deleted account should not continue starting client-owned remote operations. An already transmitted server request cannot be guaranteed to be undone, but the client can stop queued work and propagate best-effort abort to active fetches.

Node's `AbortController` is the native cancellation mechanism for Promise/network APIs. Mature queue implementations similarly remove aborted waiting tasks from the queue and require already-running tasks to observe the same AbortSignal.

## Scope

- retain the current global concurrency limit and translation request/result contracts;
- tag queued remote work with its partition owner;
- on account deletion, reject not-yet-started queue entries owned by that partition before they can call `fetchImpl`;
- track active AbortControllers per partition and abort only that partition's running fetches;
- make the remote task re-check deleted ownership before beginning an attempt, closing the dequeue-to-controller race;
- classify deletion cancellation as `翻译账号已删除`, not a gateway timeout/failure retry;
- preserve other accounts' queued/running translations;
- no Translation Worker/D1/quota schema/provider semantics, UI, version, release marker or deployment change.

## Validation

Add an executable runtime contract that proves:
1. fill the 20-slot remote concurrency pool with account A work and queue at least one additional A request;
2. deleting A aborts every active A signal and rejects the queued A request without starting its fetch;
3. all A requests settle as account-deleted rather than retrying another gateway;
4. a concurrent B request remains runnable/unaffected;
5. current cache-isolation and IPC ownership contracts remain green;
6. exact-head repository tests, Cloudflare validation and Electron E2E pass before merge.

9529360-cpu:
Closes #397.

## Root cause

`src/translation-runtime.cjs::deleteAccount(partition)` only cleared account-scoped cache/inflight indexes and marked the partition deleted. The actual remote work was not owned by that lifecycle boundary: queued tasks had no partition metadata and active `fetchImpl(..., { signal })` AbortControllers were not tracked.

That meant deleting account A could leave A's queued translations starting later and A's active requests running until completion/timeout. Clearing `state.inflight` removed only the lookup handle; it did not cancel the Promise or network I/O.

Node's native cancellation model uses `AbortController`/`AbortSignal` for running Promise/network work. Mature async queues similarly remove aborted waiting tasks from the queue while requiring already-running operations to observe the cancellation signal.

## Change

- tag each remote queue entry with its partition owner;
- reject deleted-partition queue entries before they can start `fetchImpl`;
- track active translation AbortControllers per partition;
- on account deletion, remove/reject only that partition's queued entries and abort only that partition's active fetches;
- re-check deleted ownership both after async preflight and after dequeue, closing pre-queue and dequeue-to-controller races;
- classify deletion cancellation as `TRANSLATION_ACCOUNT_DELETED` / `翻译账号已删除` instead of gateway timeout/failure, so cancellation does not poison endpoint health;
- preserve the global 20-request concurrency limit and other-account work.

An already transmitted server request cannot be guaranteed to be undone; this change closes the client-owned queue/fetch lifecycle and prevents new queued work from starting after deletion.

## Regression coverage

New focused runtime contract:

1. fills all 20 remote slots with account A requests;
2. queues A request 21 plus account B behind them;
3. deletes A and proves all 20 active A signals are aborted;
4. proves queued A never starts a 21st fetch and all A promises settle as account-deleted;
5. proves deletion does not increment gateway failure health;
6. proves B starts once A releases capacity and completes normally;
7. proves new A work after deletion fails before reaching remote fetch.

Focused local validation passed:

- `node --check src/translation-runtime.cjs`
- `node --check test/translation-account-deletion-cancel-contract.cjs`
- `node test/translation-account-deletion-cancel-contract.cjs` -> `TRANSLATION_ACCOUNT_DELETION_CANCEL_CONTRACT_OK`

GitHub branch blobs match the locally executed files exactly:

- `src/translation-runtime.cjs`: `fd2067af18b5fa2344fd6fa65b2ac0217d83b2ce`
- `test/translation-account-deletion-cancel-contract.cjs`: `8756539dca314614657a895f742c1cce6b7ad933`

## Collision boundary

Branch was cut from live master `960f57779d1ea5ceb95d91d4c1d7183774de511b`; pre-PR compare is `behind=0`. Changed files are exactly:

- `src/translation-runtime.cjs`
- `test/translation-account-deletion-cancel-contract.cjs`

This is separate from #387's Translation Worker/D1 reservation ownership and from #395's desktop Subscription mutation queue. No Worker, D1, Subscription store, UI, package, version, release marker, or deployment files are changed.

Require exact-head repository tests, Cloudflare validation and Electron E2E before merge.


9529360-cpu:
Closing as already integrated by later master history. Both owned final blobs are already byte-identical on current `master`:

- `src/translation-runtime.cjs` = `fd2067af18b5fa2344fd6fa65b2ac0217d83b2ce`
- `test/translation-account-deletion-cancel-contract.cjs` = `8756539dca314614657a895f742c1cce6b7ad933`

The account-deletion cancellation semantics and regression coverage are already present. Merging this stale ancestry would add no source content.

PR #404, opened 2026-09-13
fix: serialize internal CDP flows per partition

### LINKED ISSUE #402: [P1] Serialize internal CDP flows per account partition
## Root cause

`src/internal-cdp.cjs::run()` owns attach/send/listener/detach around one matched guest, but concurrent calls for the same account/guest are not serialized.

A concrete race is:

1. run A sees `debugger.isAttached() === false`, attaches, and enters its callback;
2. run B starts while A is active, sees the same debugger already attached, and reuses that session without owning the attach;
3. A finishes first and, because A originally attached the debugger, executes `detach()` in `finally`;
4. B is still inside its callback and subsequent `sendCommand()` calls now run against a detached debugger.

Current callers include WhatsApp file sending, generic CDP attachment injection and Telegram native attachments, so separate UI/background operations can contend for the same account guest.

Electron exposes attach/detach as lifecycle operations on a specific WebContents Debugger session. Mature async mutex implementations serialize access to a shared resource with `runExclusive`, while keyed locks preserve concurrency for different resources.

## Scope

- add a small zero-dependency keyed execution queue inside the existing internal-CDP owner;
- serialize complete `run()` lifecycles by account partition so attach/callback/listener cleanup/detach cannot overlap for the same account;
- resolve the guest at execution time after the partition lock is acquired;
- preserve parallel CDP operations for different partitions/accounts;
- preserve preferred guest ID matching, external-CDP fail-closed policy, command timeout, listener cleanup and attach/detach ownership semantics;
- failed runs must release the queue so the next same-partition operation can proceed;
- no main caller, attachment transport, external-CDP policy, WebView security, package/version/release change.

## Validation

Extend/add focused executable coverage proving:
1. two same-partition runs never overlap callbacks;
2. run B starts only after A has detached, attaches its own fresh session, and can send successfully;
3. a failed first run still releases the same-partition queue;
4. different partitions can execute concurrently;
5. existing guest-selection, timeout, event-listener and external-debugging contracts remain green;

9529360-cpu:
Closes #402.

## Root cause

`src/internal-cdp.cjs::run()` owns one matched guest's debugger attach/callback/listener-cleanup/detach lifecycle, but concurrent runs for the same account were not serialized.

A concrete race was possible:

1. run A saw `debugger.isAttached() === false`, attached, and entered its callback;
2. run B began while A was active, saw the same debugger already attached, and reused it without owning that attach;
3. A finished first and detached in its `finally` because A originally attached the debugger;
4. B was still running and its later `sendCommand()` could hit a detached debugger.

Current callers include WhatsApp file sending, generic CDP attachment injection and Telegram native attachments. In addition to debugger attach ownership, those flows can manipulate page/file-picker state and should not interleave on one account guest.

Electron exposes `attach()` / `detach()` as lifecycle operations on one WebContents Debugger session. Mature async mutex implementations use exclusive execution for a shared resource; a keyed queue preserves concurrency between distinct resources.

## Change

- add one zero-dependency `partitionRuns` keyed execution queue inside the existing internal-CDP owner;
- serialize the complete `run()` lifecycle by account partition;
- resolve the guest only after the partition lock is acquired, so a queued operation uses the current guest at execution time;
- preserve parallel CDP operations for different partitions/accounts;
- preserve preferred guest matching, external-debugging fail-closed behavior, command timeout, event listener cleanup and original attach/detach ownership semantics;
- a rejected run releases the queue tail so the next same-partition operation can proceed.

No caller, attachment transport, external-CDP policy, WebView security, dependency, package/version or release behavior changes.

## Regression coverage

New focused contract proves:

- two same-partition runs never overlap their callbacks;
- run B begins only after A fully detaches, then attaches its own fresh debugger session and sends successfully;
- a failed first run releases the queue so a subsequent run succeeds;
- different partitions still execute concurrently rather than sharing a global lock.

Focused local validation passed on the exact files pushed to GitHub:

- `node --check src/internal-cdp.cjs`
- `node --check test/internal-cdp-concurrency-contract.cjs`
- `node test/internal-cdp-concurrency-contract.cjs` -> `INTERNAL_CDP_CONCURRENCY_CONTRACT_OK`

GitHub content blobs match the locally executed files exactly:

- `src/internal-cdp.cjs`: `db3e996b66a7f7f74c45655a8fdbb1a01f0e820d`
- `test/internal-cdp-concurrency-contract.cjs`: `a7c84edf040ccc4582eb54870fac7bcd558d1b83`

## Collision boundary

Branch was cut from live master `a98b967cf40c214da4ec550d5323c54d47601d04`. Fresh pre-PR compare is `behind=0`, and changed files are exactly:

- `src/internal-cdp.cjs`
- `test/internal-cdp-concurrency-contract.cjs`

Open external-CDP PR #372 owns `src/external-debugging-policy.cjs` and its contract, not this owner. Translation #401, Subscription/logout work and Broadcast #390/#396 are file-disjoint.

Require exact-head repository tests and Electron E2E before merge.


9529360-cpu:
Closing as already integrated by later master history. Both final blobs owned by this PR are already byte-identical on current `master`:

- `src/internal-cdp.cjs` = `db3e996b66a7f7f74c45655a8fdbb1a01f0e820d`
- `test/internal-cdp-concurrency-contract.cjs` = `a7c84edf040ccc4582eb54870fac7bcd558d1b83`

Per-partition CDP serialization and the concurrency regression coverage are therefore already present; merging this stale ancestry would add no source content.

PR #414, opened 2026-09-13
fix: defer account child cleanup until delete commit

### LINKED ISSUE #413: [P1] Defer scheduled attachment cleanup until account deletion commits
## Root cause

`src/account-data-boundary.cjs::runAccountRemoval()` currently executes `beforeAccountRemove(...)` before calling the authoritative account removal implementation.

The live main-process wiring passes `scheduledAttachmentBoundary.cleanupAccount(accountId)` as that callback. `cleanupAccount()` durably removes the account's scheduled-broadcast attachment refs from `scheduled-broadcast-attachments.json`.

`accountState.remove(accountId)`, however, is transactional: it builds a candidate, performs the durable accounts-state write, and only then publishes the new in-memory state. If that durable write fails, the account remains present.

The current order can therefore produce a partial delete:

1. scheduled attachment refs are durably removed;
2. `accountState.remove()` fails before committing;
3. `runAccountRemoval()` detects that the account still exists and correctly cancels the account-data delete tombstone;
4. the account survives, but its scheduled attachment refs are already gone.

That can break pending scheduled broadcasts even though the user-visible account deletion failed.

Mature transaction guidance uses after-commit hooks for external/secondary irreversible effects. Rails documents file deletion after commit specifically to avoid deleting the file when the parent transaction rolls back; Django similarly recommends delaying side effects until commit rather than doing them first and attempting rollback.

## Scope

- preserve account-data `beginDelete/cancelDelete/finalizeDelete` semantics;
- preserve the existing `beforeAccountRemove` option surface for compatibility, but treat this callback as committed-removal cleanup rather than pre-commit work;
- do not run scheduled attachment cleanup while the authoritative account still exists;
- if `removeImplementation` fails before the account-state commit, cancel the tombstone and propagate the error with zero attachment cleanup;
- if the account-state commit succeeded, run the cleanup exactly once before finalizing the account-data store;
- if a later post-commit cleanup step fails, keep the existing `{ ok: true, deleted: true, cleanupPending: true }` semantics and never pretend the account was rolled back;
- no account-state file/schema change, no ScheduledAttachment Store format change, no renderer/UI, WebView, Translation, Subscription, Worker, version or release-marker change.

## Validation

Extend the executable account removal commit-semantics contract to prove:
1. a pre-commit removal failure leaves the account present and never invokes scheduled attachment cleanup;
2. a successful committed removal invokes cleanup after the parent removal implementation, not before;
3. a post-commit removal failure still invokes committed cleanup once, finalizes the tombstone, and returns cleanupPending;
4. cleanup failure after a committed parent deletion is not retried or mistaken for rollback;
5. existing account-data store and account removal contracts remain green;

9529360-cpu:
Closes #413.

## Root cause

`src/account-data-boundary.cjs::runAccountRemoval()` ran the live `beforeAccountRemove` callback before the authoritative account removal implementation.

Main wires that callback to `scheduledAttachmentBoundary.cleanupAccount(accountId)`, which durably removes scheduled-broadcast attachment refs. `accountState.remove(accountId)`, however, only commits the account deletion after its durable accounts-state write succeeds.

If the account-state write failed, the account remained present and the account-data tombstone was correctly cancelled, but its scheduled attachment refs had already been deleted. A failed account deletion could therefore partially delete child state and break pending scheduled broadcasts.

This is the same class of ordering bug addressed by mature after-commit patterns: irreversible child/external cleanup belongs after the parent commit, not before it with rollback fiction.

## Change

- keep the existing `beforeAccountRemove` option name for compatibility with current main-process wiring, but treat it internally as committed-account cleanup;
- run the authoritative removal implementation first;
- if it fails while the account still exists, cancel the account-data tombstone and propagate with **no child cleanup**;
- if it fails after the account disappeared, run committed cleanup once, finalize account-data state, and preserve `{ ok: true, deleted: true, cleanupPending: true }` post-commit semantics;
- on normal success, run committed cleanup after the parent removal and before finalizing;
- if committed cleanup itself fails, finalize rather than pretending the parent deletion rolled back, and report cleanupPending;
- update the static IPC ownership oracle that previously hard-coded the old unsafe `begin -> cleanup -> parent delete -> finalize` ordering.

No account-state schema/file format, ScheduledAttachment Store format, renderer/UI, WebView, Translation, Subscription, Worker, version, release marker, or deployment changes.

## Regression coverage

The executable removal contract proves exact call ordering for:

1. pre-commit parent failure: `begin -> parent -> cancel`, with zero scheduled cleanup;
2. successful parent commit: `begin -> parent commit -> cleanup -> finalize`;
3. parent throws after commit: committed cleanup still runs exactly once, then finalize + cleanupPending;
4. committed child cleanup failure: no retry/rollback fiction; finalize + cleanupPending.

`test/ipc-registration-ownership-contract.cjs` now locks the same production ownership order instead of enforcing the historical unsafe order.

## Validation

Exact head `3aec9c0a57e353705124680d2d781597b4cec284`:

- repository `test` #1207: success;
- `cloudflare-worker-validation` #138: success;
- `electron-e2e` #356: success;
- validation-client-build: skipped by path as expected.

The initial exact-head run correctly failed only because the static IPC ownership oracle still encoded the old unsafe ordering; the executable removal contract and Electron E2E were already green. The oracle was then updated to the new authoritative order and the full exact-head suite passed.

## Changed files / collision boundary

Branch starts from master `a6b102148df728a59ad17bd3f0de24640edf3393`. Current changed files are exactly:

- `src/account-data-boundary.cjs`
- `test/account-remove-commit-semantics-contract.cjs`
- `test/ipc-registration-ownership-contract.cjs`

Updater/payment, gateway, Translation, WebView, internal-CDP, Broadcast Job and Broadcast File Registry work are file-disjoint.

No production deploy or formal client release should be triggered by this desktop-only maintenance change.


9529360-cpu:
Closing as already integrated by later master history. The PR's three owned files are byte-identical to current `master`:

- `src/account-data-boundary.cjs` = `17fee81b90b64c818f3223ba433c229e98bcf099`
- `test/account-remove-commit-semantics-contract.cjs` = `b4fa0b1e4933934a4a7c4a7a01db159190b6c1c2`
- `test/ipc-registration-ownership-contract.cjs` = `1c3f70e6abe1339cd930e3bcbaf1faca98ad13e6`

The account-delete post-commit cleanup behavior and its regression coverage are already present on master; merging this stale ancestry would add no content.

PR #420, opened 2026-09-13
fix: redact userinfo from malformed logged URLs

### LINKED ISSUE #419: [P1] Redact URL userinfo on malformed log URLs
## Root cause

Geek has paired main/renderer URL log sanitizers (`src/log-url.cjs` and `ui/log-url.js`). Valid URLs are parsed with the WHATWG `URL` API and rebuilt from protocol/host/path, which naturally omits username/password. The parse-failure fallback, however, only strips query/hash and truncates:

`raw.split(/[?#]/, 1)[0].slice(0, 512)`

A malformed URL containing userinfo can therefore bypass the normal parser and write credentials into console/in-memory diagnostic logs, for example:

`https://alice:secret@example.com:bad/path?token=x`

The query is removed but `alice:secret@` survives.

OWASP logging guidance says authentication passwords, access tokens and other primary secrets should be removed/masked/sanitized rather than recorded directly. The fallback path must fail closed too, not only the normal parser path.

## Scope

- preserve current valid-URL output exactly (`protocol + //host + pathname`);
- in the parse-failure fallback, strip query/hash first, then replace scheme-authority userinfo (`scheme://...@`) with a fixed `[REDACTED]@` marker;
- keep the 512-character bound;
- apply identical semantics to `src/log-url.cjs` and `ui/log-url.js` so main and renderer logging stay in parity;
- do not attempt broad URL repair or decode/re-encode malformed input;
- no WebView navigation policy, network request, account/session, diagnostics file format, UI behavior, package/version/release change.

## Validation

Extend existing URL-redaction/parity contracts to prove:
1. valid HTTPS/chrome-extension/about URLs remain byte-for-byte unchanged from current sanitized output;
2. malformed URL userinfo never exposes username or password;
3. malformed query/hash secrets remain stripped;
4. renderer and main sanitizers return identical results for malformed userinfo cases;
5. unrelated malformed non-URL strings preserve the existing bounded fallback behavior;

9529360-cpu:
Closes #419.

## Root cause

Geek's paired main/renderer URL log sanitizers correctly omit username/password for valid URLs by parsing with `new URL()` and rebuilding only protocol/host/path. The parse-failure fallback only stripped query/hash and truncated, so malformed URLs could bypass the normal parser and leak URL userinfo into console/in-memory diagnostics.

Example before this patch:

`https://alice:secret@example.com:bad/path?token=x`

The query was removed, but `alice:secret@` survived.

OWASP logging guidance says authentication passwords, access tokens and other secrets should be removed/masked/sanitized before entering logs. The malformed fallback must therefore fail closed too.

## Change

- preserve valid-URL output exactly;
- on parse failure, strip query/hash first;
- redact `scheme://userinfo@` to `scheme://[REDACTED]@`;
- retain the existing 512-character fallback bound;
- apply identical semantics to `src/log-url.cjs` and `ui/log-url.js`;
- do not attempt broad URL repair or decode/re-encode malformed input.

No WebView navigation policy, network request behavior, account/session state, diagnostics file format, dependency, package/version or release change.

## Regression coverage

Existing main and renderer redaction contracts now additionally prove:

- malformed userinfo does not expose username/password;
- malformed query/hash secrets remain stripped;
- browser-executed renderer sanitizer matches CommonJS behavior;
- main and renderer sanitizers remain in parity;
- ordinary malformed non-URL fallback behavior stays unchanged and bounded.

Focused dynamic parity probe passed:

`MALFORMED_URL_REDACTION_PROBE_OK`

Exact branch blobs:

- `src/log-url.cjs`: `a5e1caf900add275651712cd9af61e36fd79e4bb`
- `ui/log-url.js`: `71b55fc7fda0c761e59fc00f23836e3284010e87`
- `test/webview-log-url-redaction-contract.cjs`: `72fc545fc1bf66078ae7bb6413b1055f0a7a7532`
- `test/renderer-webview-log-url-redaction-contract.cjs`: `e195b5096673d913a1e45a78ca9e92edae8fa95b`

Branch starts from live master `a6b102148df728a59ad17bd3f0de24640edf3393`; fresh compare is `behind=0`. Changed files are exactly the two URL sanitizer helpers and their two existing contracts.

Historical `agent/redact-*webview-log-urls` branches have no open PR and represent the earlier introduction of the current sanitizer owner; this issue is a newly identified malformed-userinfo edge. Current Broadcast/account/gateway/updater/payment/Translation lanes are file-disjoint.

Require exact-head repository tests and Electron E2E before merge.


9529360-cpu:
Closing as already integrated by later master history. All four exact final blobs recorded by this PR are already present on current `master`:

- `src/log-url.cjs` = `a5e1caf900add275651712cd9af61e36fd79e4bb`
- `ui/log-url.js` = `71b55fc7fda0c761e59fc00f23836e3284010e87`
- `test/webview-log-url-redaction-contract.cjs` = `72fc545fc1bf66078ae7bb6413b1055f0a7a7532`
- `test/renderer-webview-log-url-redaction-contract.cjs` = `e195b5096673d913a1e45a78ca9e92edae8fa95b`

The branch is stale only by ancestry; refreshing/merging it would add no product or contract content and would make the integration queue noisier.

PR #421, opened 2026-09-13
fix: verify payment refresh against the current order

### LINKED ISSUE #407: [P1] Verify desktop payment refresh against the current order, not account quota
## Root cause

The desktop subscription payment view currently treats `refresh().remaining_chars > 0` as proof that the just-created order was paid:

```js
const state = await window.api.subscription.refresh();
if (state.valid && state.remaining_chars > 0) {
  // “字符余额已更新”
}
```

That is not an order-state check. New accounts already receive 20,000 free characters, so a user with a still-`pending` manual order can click “我已完成付款，刷新状态” and immediately receive a false success message solely because the account already had quota.

The server already exposes authenticated `GET /api/orders`, with order statuses `pending / processing / paid / cancelled / expired`, and the desktop store already has `myOrders()`. The website already follows the correct pattern: retain the created `orderId`, query `/api/orders`, find that exact order, and only treat `order.status === 'paid'` as payment confirmation.

External payment/order guidance follows the same boundary: customer-facing status polling should query the specific order/payment identifier and finalize only from that order's reconciled status, not infer payment from an account-level balance.

## Scope

- no Subscription Worker, D1 schema, pricing, pay-method, confirmation, quota or entitlement semantics change;
- no `src/subscription.cjs` change;
- expose one trusted desktop IPC/preload method that accepts a positive integer order ID, calls existing `store.myOrders()`, and returns only a minimal `{ id, status }` projection for that authenticated user's matching order;
- do not expose the full order list or payment metadata to the renderer;
- retain the order ID returned by `createOrder()` in the payment view;
- “刷新状态” must query that exact order first;
- only `paid` may show payment success and then refresh/display quota;
- `pending`/`processing` remain unresolved, `cancelled`/`expired` are terminal failures, missing/unknown states fail closed;
- existing free quota must never be interpreted as proof that the current order was paid;
- no main/WA/LINE/account-context/Broadcast/translation/updater/version/release-marker/deployment changes.

## Validation

Add executable contracts proving:
1. the IPC boundary validates a positive safe integer order ID, remains trusted-sender protected, calls existing `myOrders()`, and returns only `{id,status}` for the matching order;
2. a newly created pending order plus `remaining_chars: 20000` does **not** render payment success and does not use quota as payment evidence;
3. `paid` for the same order triggers quota refresh and the existing success/reload flow;
4. another paid order cannot satisfy the current pending order;
5. cancelled/expired/missing/unknown states do not render success;

Reference pattern already present in this repository: the website polls `/api/orders`, finds the current `orderId`, and only accepts `order.status === 'paid'`.

9529360-cpu:
Closes #407.

## Root cause

The desktop payment flow treated account quota as proof that the newly-created order had been paid. New accounts already receive free characters, so a pending manual order could incorrectly show payment success after `subscription.refresh()` returned a positive balance.

## Change

- retain the exact order id returned by `createOrder()` in the desktop subscription UI;
- add one trusted `subscription:get-order-status` IPC/preload capability;
- validate a positive safe integer order id before touching the store;
- reuse the authenticated `store.myOrders()` result and expose only `{ id, status }` to the renderer;
- fail closed for missing or unknown order state;
- only an exact `paid` status may trigger the subsequent quota refresh and success UI;
- keep `pending` / `processing` unresolved and `cancelled` / `expired` terminal;
- never use free or cached account quota as payment proof.

No Subscription Worker, D1 schema, pricing, payment method, entitlement semantics, main process ownership, WebView, Translation, Broadcast, updater, version, release marker, or deployment changes.

The branch is rebased-by-merge onto live master `a6b102148df728a59ad17bd3f0de24640edf3393` after the subscription session-generation hardening landed. Final compare is `behind=0` and touches exactly five files:

- `src/preload.cjs`
- `src/subscription-ipc.cjs`
- `test/subscription-ipc-owner-contract.cjs`
- `test/subscription-ipc-production-ownership-contract.cjs`
- `ui/subscription.html`

It does not modify `src/subscription.cjs`, so the newly merged generation/async-completion protections remain owned by master.

## Validation

Focused contract validation performed during implementation covered:

1. trusted sender gating before store access;
2. positive safe integer order id validation;
3. minimal `{ id, status }` projection only;
4. exact pending order cannot be satisfied by another paid order or by a 20,000-character free quota;
5. exact paid order is the only path that refreshes quota and reports success;
6. cancelled, expired, missing, and unknown states fail closed.


9529360-cpu:
Closing as already integrated by later master history. All six files in the branch delta are byte-identical to current `master`: `src/preload.cjs`, `src/subscription-ipc.cjs`, `test/account-identity-contract.cjs`, both subscription IPC ownership contracts, and `ui/subscription.html`. The payment flow already requires the exact created order to reach `paid`; this stale-ancestry PR would add no source content.

Issue #21, opened 2026-08-18
ops: Cloudflare control-plane and deployment status


9529360-cpu:
Persistent non-sensitive status channel for GitHub → Cloudflare infrastructure verification and production Worker deployment reporting.

Current publishers:

- `verify-cloudflare-infra`: read-only control-plane capability/status checks;
- `deploy-website`: tests, deploys, verifies `https://geek.bbnba.com/health`, then reports here;
- `deploy-translate`: tests, deploys, verifies the translation `/health` endpoint, then reports here;
- `deploy-release-worker`: tests, deploys, verifies the public updater `latest.yml`, then reports here;
- `deploy-subscription`: tests, deploys, verifies `https://admin.bbnba.com/health`, includes the production account-smoke outcome, then reports here. The detailed transient registration/login smoke also remains in #23.

Each deploy workflow reports directly from the same GitHub-hosted job that performs the Wrangler deployment. A local maintenance container's DNS is not used as production evidence.

Allowed report fields are limited to service/resource names, pass/fail state, HTTP status codes, public endpoint names, workflow run identifiers/links, commit SHA and timestamps.

Never post API token values, Authorization headers, cookies, provider secrets, JWTs, passwords, D1/R2/KV contents, DNS record values, private Cloudflare response bodies, account data or user data.

This issue exists so future maintenance sessions can verify Cloudflare deployment/control-plane state through GitHub without depending on direct Cloudflare dashboard access.

9529360-cpu:
Run: https://github.com/9529360-cpu/geek/actions/runs/32299699558

### Client updater publication

- Service: `release artifacts / updater metadata`
- Workflow: `release-client`
- Trigger: `push` (`.github/release-client-version`)
- Version: `1.2.11`
- Commit: `de0ce48ef1ad569b32fe590a1edf51fe028592a2`
- Windows tested build: **success** (70/70 contracts inside `npm run dist`)
- Artifact validation: **success** (`geek-setup-1.2.11.exe`, blockmap, `latest.yml`, release manifest)
- Previous stable / rollback target: `1.2.10` verified
- Immutable installer + blockmap upload: **success**
- Rollback metadata snapshot: **success**
- `latest.yml` publication: **success**, published after immutable artifacts
- Public updater verification: **success** (`public-release=1.2.11`)
- Overall: **success**

[...]

PR #424, opened 2026-09-13
fix: reconcile ambiguous translation quota reservations

### LINKED ISSUE #387: [P1] Make translation quota reservation recoverable after unknown D1 outcomes

9529360-cpu:
Closes #387.

## Root cause

The translation Worker already uses `X-Request-ID` as its durable request identity, but the source-character reservation treated any thrown D1 batch error as a duplicate request.

That is unsafe for an unknown transaction outcome: if D1 committed the `translation_usage` row and source-quota debit but the caller lost the response, the Worker returned 409, left the user's quota debited, and stranded the row in `reserved`. A retry with the same request ID was then blocked forever.

D1 batches are transactional, but a caller-side exception does not prove that COMMIT did not happen. The reservation therefore needs attempt ownership plus durable reconciliation rather than inference from the exception.

## Change

- reuse the existing `translation_usage.status` text field as an attempt-owned marker `reserved:<uuid>`; no schema migration;
- insert a reservation only when the authenticated user is active and has enough quota;
- debit source quota in the same D1 batch only when the exact request/user/chars/owner marker exists;
- if the D1 batch itself throws, read the durable request row:
  - exact owner marker + user + chars => this attempt committed and may continue;
  - a different durable row => real duplicate request;
  - no durable row => propagate the storage failure instead of mislabeling it as duplicate;
- keep post-batch invariant failures outside the ambiguous-outcome reconciliation path, so a known bad debit result cannot be promoted to a successful reservation;
- bind `refundUsage` and `finishUsage` to the exact owner marker so one concurrent attempt cannot mutate another attempt's reservation;
- successful rows still transition to the existing terminal `complete` state.

Source+target charging, JWT/rate limits, provider selection, request-ID format, Subscription/payment/UI behavior and D1 schema remain unchanged.

## Executable regression coverage

The real Worker + `node:sqlite` transactional D1 contract now proves:

1. normal success still charges 5 source + 4 target characters and ends `complete`;
2. provider failure refunds only the owned reservation, removes its row, and permits safe same-ID retry;
3. reservation COMMIT followed by a simulated caller error is reconciled from the exact owner marker, reaches exactly one provider call, and completes without a double debit;
4. while request A owns a reservation, a concurrent same-request-ID attempt gets 409, makes no provider call, and cannot replace/refund/finish A's marker;
5. insufficient quota creates no usage row and calls no provider;
6. simulated pre-COMMIT D1 failure returns `translation_failed`, leaves quota unchanged, and is not mislabeled as `duplicate_request`;
7. the existing finish-COMMIT-then-error regression remains covered so catch/refund cannot mint quota after a terminal completion.

The concurrency probe fails fast if a duplicate request ever reaches the provider, avoiding a false-green/hung test.

Two older static authority/auth contracts initially exposed stale assumptions that the balance condition had to appear in the historical one-line `UPDATE`. They now verify the current stronger invariant instead: the same reservation batch gates row creation on sufficient active-user quota and binds the debit to the exact owner row. No single-writer/auth requirement was relaxed.

The historical one-commit branch was not replayed wholesale. Its Worker change was merged onto live master `a6b102148df728a59ad17bd3f0de24640edf3393` without force-push, then current contracts were updated. Final compare at head `05838c6df80fcab838f4f050107a99de74b69f04` is `behind=0` and changes exactly:

- `scripts/geek-translate-worker.js`
- `test/translation-usage-state-contract.cjs`
- `test/subscription-legacy-usage-authority-contract.cjs`
- `test/translation-gateway-auth-contract.cjs`

It is file-disjoint from desktop Translation cancellation #401 and the current Subscription/WebView/Broadcast/updater/payment lanes.

Exact pushed blob identities used for focused validation:

- Worker: `c7f3c319a763388a05541f6b470ad6cd32f34aca`
- usage-state contract: `062cd0a5055000f1bad1e063d4c20f9c72455c86`
- legacy quota-authority contract: `5a70626811d01d3f2791a519e5a56f9bca66ff63`
- gateway-auth contract: `11ef99aa3c922f286dcce211eb2dd77acfe3410e`

Focused local validation included syntax checks plus `TRANSLATION_USAGE_STATE_CONTRACT_OK`, repeated three times to exercise the concurrent duplicate path without flake.

## Exact-head CI

All required gates are green on head `05838c6df80fcab838f4f050107a99de74b69f04`:

- repository `test` run `34780802296` — success;
- `cloudflare-worker-validation` run `34780802302` — success, including repository tests, Worker/deployment-tooling syntax, and dry-run bundles for website/subscription/translation/release Workers;
- `electron-e2e` run `34780802499` — success, including diff whitespace, Linux Electron shell smoke with Secret Service, and Windows WhatsApp cold-start gate;
- `validation-client-build` is correctly skipped because this Worker-only change does not touch client-build paths.

## Delivery gate

This PR changes the Translation Worker. Merging can trigger production Translation deployment. The implementation is validated and review-ready, but **do not merge without explicit production authorization**.


9529360-cpu:
2026-09-14 live-state reconciliation: this PR is no longer a pending production change.

Current `master` already contains the same quota-reconciliation implementation via release integration #434 / merge commit `4901d872f47e1472a3d373e386bac705db41e220`:

- `scripts/geek-translate-worker.js` blob on current master is exactly `c7f3c319a763388a05541f6b470ad6cd32f34aca` (same as this PR);
- `test/translation-usage-state-contract.cjs` is exactly `062cd0a5055000f1bad1e063d4c20f9c72455c86`;
- `test/subscription-legacy-usage-authority-contract.cjs` is exactly `5a70626811d01d3f2791a519e5a56f9bca66ff63`;
- `test/translation-gateway-auth-contract.cjs` contains this PR's contract and has since been strengthened with checkout-EOL normalization.

Production Translation deployment for merge commit `4901d872f47e...` is recorded on #21 by `deploy-translate` run `34782441045`: Wrangler deploy success and public `/health` verification success (HTTP 200). Issue #387 is already closed as completed.

Closing this PR as superseded/absorbed by current master rather than refreshing or merging a duplicate Worker diff. Live GitHub remains the authority for future sessions.

PR #425, opened 2026-09-13
fix: retain newest diagnostics rotation files

### LINKED ISSUE #423: [P2] Preserve newest diagnostics files after double-digit rotation
## Root cause

`src/diagnostics.cjs::cleanupOldFiles()` currently does:

1. list every `*.jsonl` file in the diagnostics directory;
2. lexicographically `.sort()` the filenames;
3. repeatedly delete the first filename until `maxFiles` remains.

Geek names rotated files as `diagnostics-${timestamp}-${fileIndex}.jsonl`, where `fileIndex` is not zero-padded. Once one timestamp bucket reaches a two-digit index, lexicographic order no longer matches generation order: `...-10.jsonl` sorts before `...-7.jsonl`.

Deterministic current-algorithm proof with indexes `0..14` and `maxFiles=3`:

- retained by current cleanup: `7, 8, 9`
- actual newest generations: `12, 13, 14`

That means high-frequency diagnostics can delete the newest incident evidence while retaining older files.

Mature rotating-log implementations keep explicit/audited rotation history or otherwise order generations by real rotation identity rather than arbitrary filename string order. `winston-daily-rotate-file`, for example, tracks retained log files through an audit file for `maxFiles` cleanup.

## Scope

- keep the existing file naming format for compatibility;
- order Geek-owned diagnostic files by timestamp bucket and numeric trailing generation index;
- preserve `maxFiles` behavior and best-effort logging semantics;
- do not change event payloads, privacy sanitization, log size limits, directory, package/version or release behavior;
- keep existing diagnostics files readable and eligible for cleanup without migration/rename.

## Validation

Extend the executable diagnostics contract with a deterministic fixed-time rotation case:

1. `maxBytes=1`, `maxFiles=3`, fixed `now()`;
2. write 15 entries so generations reach `-14.jsonl`;
3. prove exactly the newest generations / sequence values `12,13,14` survive;
4. preserve current privacy/redaction and max-files assertions;
5. exact-head repository tests and Electron E2E must pass before merge.

9529360-cpu:
Closes #423.

## Root cause

`src/diagnostics.cjs::cleanupOldFiles()` sorted rotated filenames lexicographically before deleting the oldest entries. Geek names files as `diagnostics-${timestamp}-${fileIndex}.jsonl`, but `fileIndex` is not zero-padded.

Once one timestamp bucket reaches a double-digit generation, filename order stops matching rotation order. Deterministic pre-fix proof with generations `0..14` and `maxFiles=3`:

- current lexicographic cleanup retains `7, 8, 9`;
- the actual newest generations are `12, 13, 14`.

High-frequency logging could therefore delete the newest incident evidence while retaining older files.

Mature rotating-log implementations track rotation/retention identity explicitly instead of relying on arbitrary filename string ordering. This patch follows that principle without adding a dependency or changing Geek's filename format.

## Change

- keep the existing `diagnostics-${timestamp}-${fileIndex}.jsonl` format;
- parse Geek-owned diagnostic filenames into `{ timestamp, numeric generation }`;
- order cleanup by timestamp bucket, then numeric generation;
- delete only the oldest parsed Geek diagnostics until `maxFiles` remains;
- preserve existing max byte/file limits, best-effort failure behavior, event payloads and privacy sanitization;
- no migration or rename of existing files.

No UI, WebView, account/session, Subscription, Translation, Broadcast, updater, dependency, package/version, release marker or deployment changes.

## Regression coverage

The existing executable diagnostics contract now adds a fixed-time rotation case:

1. `maxBytes=1`, `maxFiles=3`, fixed `now()`;
2. write 15 entries so the generation reaches `-14.jsonl`;
3. assert exactly generations `12,13,14` survive;
4. parse retained log payloads and assert their event sequence values are also exactly `12,13,14`.

The unchanged pre-fix algorithm was run against this same contract and fails deterministically with actual generations `[7,8,9]`. The patched exact files pass the full diagnostics contract.

- `node --check src/diagnostics.cjs`: pass
- `node --check test/diagnostics-contract.cjs`: pass
- `node test/diagnostics-contract.cjs` -> `DIAGNOSTICS_CONTRACT_OK`

Exact GitHub/local blobs:

- `src/diagnostics.cjs`: `e39e02a0b7d7638c2863c0c114d8d6b7f4325480`
- `test/diagnostics-contract.cjs`: `1aa9022c306559d3b5c457869c91fafd42e36af7`

Branch starts from live master `a6b102148df728a59ad17bd3f0de24640edf3393`. Final pre-PR compare is `behind=0`, with exactly two changed files:

- `src/diagnostics.cjs` (+15/-2)
- `test/diagnostics-contract.cjs` (+13/-0)

Historical `agent/diagnostics-privacy` is a stale/diverged research branch with no open PR; current master was roughly 745 commits ahead when audited. Current payment, updater, Broadcast, account deletion, gateway, Translation, WebView/internal-CDP and log-url lanes are file-disjoint.


9529360-cpu:
Superseded by current-master integration PR #436. The exact validated diagnostics blobs from this PR are carried into #436 on top of post-#434 master, where they are being revalidated together with the remaining session wrap-up fixes. Closing this feature PR prevents duplicate integration.

PR #430, opened 2026-09-13
fix: ignore stale scheduled broadcast generations

### LINKED ISSUE #428: [P1] Ignore stale scheduled-broadcast callbacks after reschedule
## Root cause

`ui/broadcast-schedule-registry.js::schedule()` cancels an existing `accountId + taskId`, stores the replacement task, and arms a new timer. That safely cancels a timer that has not fired yet, but it cannot cancel an old timer callback that has already started and is awaiting the asynchronous `beforeDue(task)` durability gate.

The old callback currently has no generation/identity check after that await and its `finally` unconditionally deletes `tasksByAccount[accountId][taskId]`.

Concrete race:

1. task generation A becomes due and its timer callback starts;
2. A awaits `beforeDue(A)`;
3. while A is suspended, the same task ID is rescheduled as generation B;
4. `schedule(B)` replaces the registry entry and arms B's timer;
5. A's `beforeDue` resolves;
6. stale A still calls `onDue(A)`;
7. A's `finally` deletes the registry entry by ID, removing B's logical task even though B is the current generation.

This can send an obsolete scheduled broadcast and corrupt the replacement schedule's registry state.

Mature async stale-completion patterns use a generation/identity guard after asynchronous boundaries: work that can no longer be physically cancelled must be prevented from mutating current state once a newer generation owns that logical key. React's official race-condition guidance uses the same ignore-stale-result principle.

## Scope

- keep the existing schedule/cancel/rearm public API and timer chunking behavior;
- treat the frozen task object stored in `tasksByAccount` as the generation token;
- after `await beforeDue(task)`, run `onDue(task)` only if the registry still points to that same task object;
- in `finally`, delete by ID only if the registry still points to that same task object;
- preserve already-started `onDue` semantics: rescheduling after `onDue` has begun cannot undo the send, but its completion must not delete a newer generation;
- no Job Manager, persistence format, attachment, renderer app.js, package/version/release/deployment changes.

## Validation

Extend the executable schedule-registry contract with a deterministic barrier:
1. generation A timer starts and blocks inside `beforeDue(A)`;
2. generation B with the same account/task ID is scheduled while A is suspended;
3. releasing A's barrier must not call A's `onDue`;
4. A's stale `finally` must not remove B, and B must remain armed;
5. advancing the clock and firing B runs B exactly once and then removes B normally;
6. existing far-future timer chunking and pre-fire cancellation behavior remain green;
7. exact-head repository tests and Electron E2E pass before merge.

9529360-cpu:
Closes #428.

## Root cause

Rescheduling the same `accountId + taskId` cancels the previous timer handle and replaces the registry entry, but an old timer callback may already have started and be suspended inside the asynchronous `beforeDue(task)` durability gate.

The old callback previously resumed with no generation check, could call `onDue(oldTask)`, and its unconditional `finally` deleted the registry entry by ID. That allowed an obsolete scheduled broadcast to run and could remove the newer replacement task from `tasksByAccount`.

Deterministic pre-fix probe reproduced the split state exactly after replacing an in-flight generation:

- stale `oldDue = 1`;
- replacement `hasNew = false`;
- replacement timer handle still physically present (`timerCount = 1`).

This is the same stale-async-completion class documented in mature UI/runtime patterns: once already-started async work cannot be physically cancelled, a generation/identity guard must prevent it from mutating current state after ownership moves to a newer generation.

## Change

- keep the existing task object as the generation token already stored in `tasksByAccount`;
- after `await beforeDue(task)`, return without `onDue` unless the registry still points to that exact task object;
- in `finally`, delete by task ID only when the registry still points to that exact task object;
- preserve schedule/cancel/rearm API, far-future timer chunking, persistence gate, callback signatures and timer policy;
- if `onDue` itself already started before a later reschedule, it is not retroactively cancelled, but its completion cannot delete the newer generation.

No Job Manager, schedule persistence format, ScheduledAttachment Store, `ui/app.js`, backend/Worker, package/version, release marker or deployment changes.

## Regression coverage

The existing executable schedule-registry contract now adds a deterministic barrier:

1. old generation A becomes due and its callback blocks inside `beforeDue(A)`;
2. replacement generation B with the same account/task ID is scheduled while A is suspended;
3. releasing A proves stale A never invokes its `onDue`;
4. stale A's `finally` cannot remove B;
5. B remains the registered task with one armed timer;
6. advancing the fake clock fires B exactly once and B then removes itself normally.

Existing 90-day timer chunking and normal pre-fire cancellation coverage remain unchanged and green.

- `node --check ui/broadcast-schedule-registry.js`: pass
- `node --check test/broadcast-schedule-registry-contract.cjs`: pass
- `node test/broadcast-schedule-registry-contract.cjs` -> `BROADCAST_SCHEDULE_REGISTRY_CONTRACT_OK`

Exact GitHub/local blobs:

- `ui/broadcast-schedule-registry.js`: `3eb034672203293aede3ef2a0f3ab6eb8f29339f`
- `test/broadcast-schedule-registry-contract.cjs`: `6790069a64f6f83be71e58c46cc571aaa268c14c`

Branch starts from live master `a6b102148df728a59ad17bd3f0de24640edf3393`. Final pre-PR compare is `behind=0`, with exactly two changed files:

- `ui/broadcast-schedule-registry.js` (+2/-1)
- `test/broadcast-schedule-registry-contract.cjs` (+45/-0)

Current Broadcast Job #390, Broadcast File Registry #416, readiness #418, diagnostics #425, account deletion #414, internal-CDP #404, Translation/Subscription/updater/payment and WebView lanes are file-disjoint.


9529360-cpu:
Superseded by the merged 1.2.23 integration PR #434. Current master contains the exact validated schedule-registry source blob `3eb034672203293aede3ef2a0f3ab6eb8f29339f` and regression blob `6790069a64f6f83be71e58c46cc571aaa268c14c`, so this feature PR should not be merged a second time.

PR #432, opened 2026-09-13
fix: bind Geek userData before early Electron boundaries

### LINKED ISSUE #431: [P1] Create and bind Geek userData before early Electron boundaries
## Root cause

`src/main-entry.cjs` resolves the profile-specific `earlyUserDataDir` and immediately does:

`try { app.setPath('userData', earlyUserDataDir); } catch {}`

before taking the single-instance lock and before installing Session/navigation/permission boundaries.

Electron's current official `app.setPath()` contract states that if the target directory does not exist, `setPath` throws and the application should create the directory first with `fs.mkdirSync` or similar.

Geek currently creates the fixed userData directory later inside `runtimePaths.migrateRuntimeFiles()` in `main.cjs`. That is too late for a clean install: `main-entry` needs the path before single-instance and Session initialization.

On a first run where the target directory is absent, the early `setPath` can fail silently. The process can then continue with Electron/Chromium using the default profile path while Geek's accounts/config owners still use the explicit resolved `USER_DATA_DIR`, splitting browser Session state from authoritative account/config files and weakening profile-scoped single-instance isolation.

## Scope

- create the resolved early userData directory recursively before `app.setPath`;
- keep the path binding before `installSingleInstanceGuard()` and every Session/WebView boundary;
- stop startup before the single-instance/main bootstrap if directory creation or `setPath` fails instead of silently falling back to Electron's default profile;
- isolate the bind operation in a small testable helper rather than duplicating filesystem behavior;
- keep current runtime profile resolution, file migration, accounts/config formats, Chromium partition names and release/version behavior unchanged.

## Validation

Add executable coverage that proves:
1. a nested/nonexistent clean-install target is created before the fake Electron App receives `setPath('userData', target)`;
2. mkdir failure propagates;
3. `setPath` failure propagates rather than being swallowed;
4. `main-entry.cjs` gates `installSingleInstanceGuard()` and `require('./main.cjs')` behind successful userData binding;
5. existing runtime-profile, runtime-paths, single-instance, Session permission/navigation and Electron E2E contracts remain green;

9529360-cpu:
Closes #431.

## Root cause

`src/main-entry.cjs` resolved the profile-specific Geek `earlyUserDataDir` and called `app.setPath('userData', earlyUserDataDir)` before single-instance and Session/WebView boundaries, but it silently swallowed any exception.

Electron's current official `app.setPath()` contract requires the target directory to already exist; otherwise `setPath` throws and Electron recommends creating the directory first. Geek only created the fixed userData directory later inside `main.cjs -> migrateRuntimeFiles()`, which is after the early profile/session boundaries need the path.

On a clean install, the early bind could therefore fail silently and let Electron/Chromium continue on its default userData path while Geek account/config owners still target the explicit resolved Geek directory. That can split Session/login state from authoritative account/config state and weaken profile-scoped single-instance isolation.

## Change

- add a small `prepareUserDataPath()` owner that validates an absolute target, recursively creates it, then calls `app.setPath('userData', target)`;
- bind the Geek userData path before `installSingleInstanceGuard()` and every Session/WebView boundary;
- propagate mkdir/setPath failures to the startup gate instead of silently falling back;
- log a bounded error category and exit with code 1 if the early profile cannot be safely bound;
- gate the single-instance/main bootstrap on successful binding;
- leave the existing later `migrateRuntimeFiles()` behavior intact for accounts/config migration.

No account/config schema, partition naming, migration format, renderer/UI, Subscription, Translation, Broadcast, Worker, dependency, package/version, release marker or deployment changes.

## Regression coverage

New executable contract proves:

1. a nested nonexistent clean-install target exists before a fake Electron App receives `setPath('userData', target)`;
2. mkdir failure propagates and `setPath` is never reached;
3. `setPath` failure propagates rather than falling back to the default profile;
4. relative targets are rejected;
5. source order remains `runtime profile -> userData prepare/bind -> single-instance -> Session boundary`;
6. startup is explicitly gated by `userDataPathReady` and exits nonzero on bind failure;
7. the historical silent `try { app.setPath(...) } catch {}` pattern is gone.

- `node --check src/user-data-path.cjs`: pass
- `node --check src/main-entry.cjs`: pass
- `node --check test/user-data-path-contract.cjs`: pass
- `node test/user-data-path-contract.cjs` -> `USER_DATA_PATH_CONTRACT_OK`

Exact GitHub/local blobs:

- `src/user-data-path.cjs`: `dc5b76945a12dd7f13da2166fe23ab5d4a2e5efb`
- `src/main-entry.cjs`: `1a36be394cdcedbd13af69629ed9f5fadcb5e5e2`
- `test/user-data-path-contract.cjs`: `caa3f3840fc91cf841e7d263fc0bd76cf973e3a8`

Branch starts from live master `a6b102148df728a59ad17bd3f0de24640edf3393`. Final pre-PR compare is `behind=0`, with exactly three changed files:

- `src/main-entry.cjs` (+11/-2)
- `src/user-data-path.cjs` (new)
- `test/user-data-path-contract.cjs` (new)

A final open-PR search found no current owner for `main-entry` / early userData binding. Existing external-CDP #372 and crash-recovery #396 explicitly own other files and do not modify `main-entry.cjs`.


9529360-cpu:
Superseded by current-master integration PR #436. This PR's exact userData helper/main-entry/test blobs (including the corrected single-instance oracle) are carried into #436 on top of post-#434 master and are being revalidated there as the single session wrap-up boundary. Closing this feature PR prevents duplicate integration.

Issue #376, opened 2026-09-13
[P2] Support updater byte ranges on the R2 release Worker for differential downloads


9529360-cpu:
## Root cause

Geek publishes updates through the generic `electron-updater` provider at `https://geek-release.9529360.workers.dev`. The release Worker currently ignores HTTP `Range` and always calls `RELEASE_BUCKET.get(key)` with no range, returning the full installer with status 200.

`electron-updater`'s differential downloader reads remote installer segments with byte-range requests. Its generic provider enables multiple range requests by default for non-S3 URLs. When the server does not answer range requests as partial content, electron-updater falls back to a full installer download. Updates still work, but the R2/blockmap differential path is effectively wasted.

Cloudflare R2's Workers API natively supports ranged reads through `get(key, { range })`, and R2 objects expose the returned range + full object size. A mature R2 public-worker pattern is to return HTTP 206 + `Content-Range` for single ranges and support HEAD metadata requests.

## Conservative implementation direction

Do not build a custom multipart/byteranges aggregator in the Worker. That would turn one public HTTP request into an attacker-controlled number of R2 reads and add substantial streaming/parser complexity.

Instead:

- make the release Worker a proper GET/HEAD static-object boundary;
- support one RFC-style `Range: bytes=...` range and return 206, `Content-Range`, `Content-Length`, and `Accept-Ranges: bytes`;
- return 416 + `Content-Range: bytes */<size>` for invalid/unsatisfiable/multi-range requests;
- preserve existing allow-list, error redaction, cache policy, ETag and content types;
- set `publish.useMultipleRangeRequest: false` in `electron-builder.yml`, which is an explicit electron-builder generic-provider option, so future released clients issue the single-range requests the Worker supports;
- current already-released clients remain safe: unsupported multi-range attempts get 416 and electron-updater falls back to its existing full-download path;
- add executable Worker behavior coverage for GET, HEAD, prefix/open-ended/suffix ranges, invalid ranges, method rejection, and no body on HEAD.

## Collision boundary

Limit changes to release Worker, future updater publish configuration, and focused release/updater contracts. Do not bump `package.json` or `.github/release-client-version`, do not trigger or claim a formal client release, and do not touch desktop runtime, LINE/WhatsApp/account-context, Subscription, or Translation Worker code.

No open release/updater/range PR exists at task start.

## Acceptance

1. GET without Range remains 200 with the full artifact.
2. HEAD returns metadata only and never reads/streams the object body.
[...]

Issue #158, opened 2026-08-22
ux: complete website onboarding and account self-service journey


9529360-cpu:
## Problem

The public website and user account center look polished, but the self-service journey is incomplete:

- footer links for “使用教程” and “常见问题” point to unrelated download/pricing sections;
- new users cannot see a concise install-to-use workflow before downloading;
- the account center does not surface account age or a clear durable-balance explanation;
- USDT payment instructions require manual text selection and provide weak copy/retry feedback;
- recent orders are terse and lack timestamps or an explicit refresh affordance.

This issue covers one root cause: the web onboarding-to-purchase journey is visually present but not operationally self-explanatory.

## Requirements

- **WEB-ONB-01**: Add real “使用流程” and “常见问题” sections to the public website and route navigation/footer links to their actual anchors.
- **WEB-ONB-02**: Keep claims within current proven capability and preserve the existing dynamic release/download path.
- **WEB-ACC-01**: Improve account summary with account creation information, clear no-expiry character-balance wording, loading/error states, and a manual refresh affordance.
- **WEB-ACC-02**: Add safe copy controls for the exact USDT amount and server-provided TRC20 address, with visible success/failure feedback and no secondary hard-coded destination.
- **WEB-ACC-03**: Improve recent-order rows with creation time, status semantics, and refresh behavior using existing authenticated APIs.
- **WEB-A11Y-01**: Preserve keyboard focus, semantic buttons/links, responsive layout, and reduced-motion behavior.
- **WEB-OBS-01**: Serve a same-origin favicon to remove the current browser 404.

## Non-goals

- no authentication, password, quota debit, payment-confirmation, D1 schema, secret, or admin API changes;
- no client version/release marker change;
- no Cloudflare credential or production-control-plane changes;
- no new externally hosted trackers, fonts, images, or payment destinations.

## Acceptance

- focused contract verifies anchors, account/payment controls, safe DOM rendering, favicon route, and dynamic download invariants;
- `npm test`, Worker syntax check, and `git diff --check` pass;
- desktop and 390 px mobile browser screenshots show no horizontal overflow and all new sections/controls are usable.

PR #436, opened 2026-09-13
integrate: close diagnostics and startup state reliability gaps

### LINKED ISSUE #423: [P2] Preserve newest diagnostics files after double-digit rotation

### LINKED ISSUE #431: [P1] Create and bind Geek userData before early Electron boundaries

### LINKED ISSUE #433: [P1] Keep missing config retryable after legacy migration I/O failure
## Root cause

Startup runs `runtimePaths.migrateRuntimeFiles()` before `configStore.load()`, but `src/main.cjs` intentionally catches migration I/O errors and continues.

For accounts, the next startup can still recover because `runtimePaths` treats an empty `accounts.json` as a replaceable first-run shadow. Config does not have that exception (`allowEmptyShadow: false`).

`src/config-state.cjs::load()` currently handles a missing primary and missing backup by immediately durable-writing `DEFAULT_CONFIG` to `config.json`.

That creates a permanent legacy-migration shadow after a transient first-run copy failure:

1. legacy `data/config.json` exists with real user preferences / proxy settings;
2. first-run `copyFile(legacy, userData/config.json)` fails (for example EBUSY/EACCES/EIO);
3. main logs the migration error and continues;
4. Config State sees no target/backup and durable-writes defaults to `userData/config.json`;
5. next startup sees the destination as existing and authoritative, so legacy config is never copied again.

Mature settings stores commonly keep defaults in memory until a real mutation is written, and mature migration runners fail/retain retryability rather than converting a migration error into an authoritative default state. `electron-store` likewise does not require schema/default values to be eagerly materialized to disk before a set operation, while its actual writes are atomic.

## Scope

- when both config primary and backup are missing, load `DEFAULT_CONFIG` into canonical memory but do not eagerly create `config.json`;
- preserve current behavior for existing primary, backup recovery, corrupt/decrypt failures, plaintext/encryption migration, update durability and transaction serialization;
- first real `update()` still creates the primary + backup atomically through the existing durability owner;
- preserve clean-install runtime behavior: callers still receive the same default snapshot;
- do not change `runtimePaths.migrateRuntimeFiles`, config schema/default values, UI, account state, package/version/release/deployment behavior.

## Validation

Add an executable retryability contract that proves:
1. with no target/backup, `load()` returns the current defaults but leaves `config.json` absent;
2. a subsequent `runtimePaths.migrateRuntimeFiles()` can still copy a legacy config into that missing target;
3. a fresh Config State instance then loads the legacy values correctly and rewrites them through the existing secure canonical migration path;
4. after a real `update()`, the config primary/backup are durably created as before;
5. existing config transaction/recovery/security contracts and #375's legacy theme migration contract remain compatible;

9529360-cpu:
## Session wrap-up integration

This is the post-release desktop reliability integration for the three file-disjoint fixes completed in the previous maintenance session. The branch has now been refreshed onto current master `0b4b66125442714366ab9864a02897e7dd01b7ca` after Geek 1.2.24 completed its production promotion and verification.

Included fixes:

- #425 / #423 — retain the newest diagnostics rotation generations using timestamp + numeric generation ordering instead of lexicographic filename order;
- #432 / #431 — create and bind the resolved Geek `userData` directory before single-instance and Session/WebView boundaries, and fail closed when that early profile cannot be bound;
- #433 — keep a missing `config.json` retryable after a transient legacy-copy failure by loading defaults in memory without eagerly creating an authoritative defaults shadow.

Closes #423
Closes #431
Closes #433

## Exact change surface

Exactly 8 files relative to current master:

- `src/diagnostics.cjs`
- `test/diagnostics-contract.cjs`
- `src/main-entry.cjs`
- `src/user-data-path.cjs` (new)
- `test/user-data-path-contract.cjs` (new)
- `test/single-instance-contract.cjs`
- `src/config-state.cjs`
- `test/config-state-missing-retry-contract.cjs` (new)

No Worker, D1, Subscription/payment, Translation, Broadcast delivery, WebView navigation/permission, dependency, package/version, formal release marker or deployment changes.

## Refresh and release boundary

Original exact head `f88977bc215b6437ccc303506b6329cfc195bdcd` was intentionally held while the release boundary was unresolved. Since then master advanced through the 1.2.24 Windows release recovery fixes. The later master changes are file-disjoint from this PR's 8-file surface.

Current exact head is `85476fcd1a2d8c10bf1a043c2249be4f66efc9a9`, created by a normal non-force merge of current master into the integration branch. No feature semantics were changed during the refresh.

Geek 1.2.24 production release-client run #26 succeeded on current master: tested Windows build, artifact validation, installer/blockmap upload, latest metadata publication, public production verification, and SHA-256 artifact integrity verification all passed. This PR therefore no longer sits inside the selected release boundary and must not trigger or modify release metadata.

## Integration gate

Require exact-head repository tests, Cloudflare validation and Electron E2E for `85476fcd1a2d8c10bf1a043c2249be4f66efc9a9` before merge. Validation client build may remain skipped by path policy because the exact 8-file surface does not select that workflow.


9529360-cpu:
## Exact-head closeout

Head `f88977bc215b6437ccc303506b6329cfc195bdcd` is integration-validated against master `4901d872f47e1472a3d373e386bac705db41e220`.

Required gates:
- repository `test` #1227: **success** — all 183 tests passed, including `CONFIG_STATE_MISSING_RETRY_CONTRACT_OK`, `DIAGNOSTICS_CONTRACT_OK`, `SINGLE_INSTANCE_CONTRACT_OK`, and `USER_DATA_PATH_CONTRACT_OK`;
- Cloudflare validation #158: **success**;
- Electron E2E #376: **success** — whitespace, Windows WhatsApp cold-start, and Linux Electron shell smoke all passed;
- validation-client-build #276: **skipped** by path policy.

Freshness check after CI: master remains `4901d872f47e1472a3d373e386bac705db41e220`; PR remains mergeable and the compare is the same one-commit/eight-file post-#434 delta.

### Intentional release hold

Do **not** merge this PR merely to clear the queue while the selected 1.2.23 candidate boundary from #434 has not yet been formally promoted. Current master still reports package version `1.2.22`. Merging #436 now would mutate the selected candidate contents before version promotion.

Once the 1.2.23 promotion/release boundary is completed (or the release owner explicitly reselects #436 into that release), refresh master and this PR's freshness. If release changes are file-disjoint, carry this exact validated delta forward as the next maintenance integration. No additional engineering work is pending on the three fixes themselves.

PR #441, opened 2026-09-13
test: make Windows release source contracts portable

9529360-cpu:
## Release blocker

Geek 1.2.24 production `release-client` run #25 reached the Windows `npm run dist` pre-build test suite and failed in `test/webview-navigation-boundary-contract.cjs`. The test located the end of the main-process `did-attach-webview` lifecycle block with an LF-only substring. On the Windows checkout, `src/main.cjs` was materialized with CRLF, so the slice became empty even though the production lifecycle instrumentation was present.

This is the same platform portability class already fixed in neighboring source-contract tests: Git may use CRLF in a Windows working tree, while source-structure tests should inspect normalized text.

## Change

- normalize `src/main-entry.cjs` and `src/main.cjs` text to LF before structural source inspection in `webview-navigation-boundary-contract.cjs`;
- preserve every WebView navigation/security assertion unchanged;
- record the exact next-release fallback history: both 1.2.23 and 1.2.24 failed before installer/blockmap/R2/latest.yml publication, so 1.2.25 must still use the real public stable 1.2.22 as website fallback;
- keep fallback exceptions exact to version pairs; do not permit arbitrary stale fallback versions.

No runtime source, WebView policy, Worker, package version, release marker, updater metadata or deployment behavior changes.

## Windows proof

A one-shot Windows validation workflow ran on the exact fix branch with Node 22 and a Windows checkout, executed the full repository `npm test`, and completed successfully. That workflow then self-removed. Final compare against master contains only the two test files in this PR.

## Release safety

The failed 1.2.24 production run stopped at `Build tested Windows release`. Artifact validation, previous-stable capture, R2 uploads, rollback metadata, `latest.yml` promotion and public verification were all skipped. Public stable therefore remains 1.2.22.

After this PR is exact-head green and merged, the next formal client release must be 1.2.25 rather than retrying an unpublished version whose source changed.


9529360-cpu:
Superseded by merged PR #442, which fixed the same Windows CRLF source-contract failure and also covered the adjacent `webview-unknown-partition` contract. The subsequent Geek 1.2.24 production release-client recovery run succeeded from the updated master, so this older release-blocker candidate is no longer actionable. Closing to keep the integration queue unambiguous.

PR #457, opened 2026-09-14
fix: make pending order admission atomic

### LINKED ISSUE #455: [P1] Make pending order admission atomic under concurrent retries
## Root cause

`POST /api/orders` in the Subscription core currently performs pending-order reuse as a read-then-insert sequence:

1. `SELECT ... WHERE user_id = ? AND plan = ? AND status = 'pending'`;
2. if no row is visible, `INSERT INTO orders ...`.

The production compatibility wrapper scopes that reuse lookup by `pay_method`, and the USDT compatibility owner already makes amount-slot allocation atomic, but there is no atomic admission boundary for "one pending order per user + plan + pay method". Two concurrent requests can both observe no pending row before either insert becomes visible, then both create pending orders. For USDT the existing amount allocator can give them different safe amounts, so amount uniqueness does not prevent duplicate logical orders.

This becomes more important after desktop request deadlines (#453/#454): aborting the client fetch does not prove the Worker/D1 write did not commit, so a manual retry can overlap an earlier unknown-outcome request.

Cloudflare D1 executes single statements transactionally and is SQLite-compatible. The repository already uses `INSERT ... SELECT ... WHERE NOT EXISTS` as the atomic USDT amount-slot admission pattern; reuse the same pattern for logical pending-order admission rather than adding a new schema/migration for this fix.

## Scope

- keep current order schema, pricing, pay methods and pending-order reuse semantics;
- make the final order insert atomically refuse a second `pending` row for the same authenticated user + plan + normalized pay method;
- preserve USDT active amount-slot uniqueness in the same atomic insert attempt;
- when a concurrent request loses the admission race, replay only the existing read/reuse path so the caller receives the committed pending order with `reuse: true`;
- do not retry provider/payment side effects and do not create a second order;
- preserve the intentional ability to have separate pending manual and USDT orders for the same plan;
- no UI, desktop state, client version/release marker, pricing/quota logic or D1 schema migration.

## Validation

Use the real compatibility wrapper + SQLite/D1 adapter contract to prove:
1. concurrent same user/plan/manual requests produce exactly one pending row and both responses identify that same order;
2. concurrent same user/plan/USDT requests produce exactly one pending row and one amount slot;
3. same user/plan with different pay methods may still create separate pending rows;
4. a pre-existing pending row still follows the current `reuse: true` path;
5. different users or plans remain independent;
6. exact-head repository tests, Cloudflare Worker validation and Electron gate remain green.

## Delivery gate

This changes the production Subscription Worker path. Implementation and CI may proceed, but merge/deploy requires explicit production authorization.

9529360-cpu:
Closes #455.

## Root cause

Subscription order creation reused an existing pending order through a read-then-insert sequence. Two concurrent `POST /api/orders` requests could both observe no pending row, then both insert. The existing USDT compatibility owner already atomically protects identifying amount slots, but that only prevents amount collisions; it does not guarantee one logical pending order per authenticated user + plan + pay method.

This is especially relevant after desktop request deadlines: client abort/timeout does not prove the Worker/D1 write did not commit, so a manual retry can overlap an earlier unknown-outcome request.

Cloudflare D1 is SQLite-compatible and the repository already uses `INSERT ... SELECT ... WHERE NOT EXISTS` as the atomic USDT amount-slot admission pattern; this PR reuses that established mechanism and adds no schema migration.

## Change

- make the final manual order INSERT atomically refuse a second `pending` row for the same user + plan + normalized pay method;
- extend the USDT atomic INSERT so logical pending-order admission and active amount-slot admission are checked in the same statement;
- preserve legacy `NULL` / empty `pay_method` as manual for compatibility;
- if another request wins the admission race, throw an internal stable `PENDING_ORDER_ALREADY_EXISTS` signal and replay only the core read/reuse path using a pre-cloned request;
- return the already committed pending order through the existing `reuse: true` behavior rather than constructing a second response model;
- keep manual and USDT pending orders independent for the same user/plan;
- keep different users/plans independent;
- keep the existing USDT 100-slot exhaustion behavior and pending/processing amount reservation unchanged;
- no D1 schema migration, pricing/quota change, UI/client state change, version/release marker change, or release workflow change.

## Executable contract

`test/subscription-order-pay-method-contract.cjs` proves:

1. the production wrapper has a bounded replay path only for the stable pending-admission conflict;
2. two stale manual prechecks followed by two inserts create exactly one pending order;
3. two stale USDT prechecks followed by two inserts create exactly one logical pending order rather than allocating a second amount;
4. manual and USDT for the same user/plan remain independent;
5. different users and plans remain independent;
6. legacy pending lookup/pay-method compatibility and USDT amount-slot behavior remain intact;
7. 100 occupied USDT amount slots still fail closed without inserting below the allowed discount floor.

## Boundary

Changed files exactly:
- `scripts/geek-subscription-worker.js`
- `scripts/subscription-order-pay-method.mjs`
- `test/subscription-order-pay-method-contract.cjs`

No `ui/`, desktop Subscription store, D1 migration, package/dependency, client version, release marker or release workflow changes.

The branch was merge-synced without force to live master `3b8310d547c718b8e8c9c721a3db52cc2b71fed4`; current exact head is `b2d9329df569fd534f06be8ace7363ffab4abe61`, with compare `behind=0` at the latest sync.

## Delivery gate

This changes the production Subscription Worker dependency closure. Exact-head repository tests, Cloudflare Worker validation and aggregate Electron E2E are required, but **do not merge this PR without explicit production authorization**, because merging to `master` matches the normal `deploy-subscription` production trigger.


9529360-cpu:
Exact-head readiness checkpoint (2026-09-14):

- head: `b2d9329df569fd534f06be8ace7363ffab4abe61`
- base: live master `3b8310d547c718b8e8c9c721a3db52cc2b71fed4`
- compare: `behind=0`; changed files remain exactly the three scoped Subscription files
- repository test workflow: success
- Cloudflare Worker validation: success, including website/subscription/translation/release dry-run bundles
- Electron E2E: success, including whitespace, Windows WhatsApp cold-start, Linux Electron shell smoke, and aggregate gate
- review threads: none
- PR is mergeable

Code/evidence status: ready candidate.

Delivery status: **blocked on explicit production authorization**. Both modified Subscription source files match the `deploy-subscription` workflow's `master` push paths, so merging this PR can trigger the production Subscription Worker deployment. Do not merge solely because CI is green.

PR #459, opened 2026-09-14
feat: make lock screen a true modal boundary

9529360-cpu:
## Summary
- make the existing lock overlay a true modal boundary without changing credential or unlock ownership
- mark the visible lock box as `role=dialog` / `aria-modal=true`, keep the background `.app` physically inert, and keep Tab/Shift+Tab inside the password + unlock controls
- keep Escape from bypassing the lock surface and redirect stray focus attempts back to the password input
- expose wrong-password feedback through `role=status`, `aria-live=polite`, `aria-invalid`, and `aria-errormessage`
- restore the pre-lock point of regard after a successful unlock
- add visible focus/error styling and reduced-motion handling
- add a real Electron E2E that locks the app, exercises focus containment, enters an intentionally wrong password, then unlocks with the actual isolated E2E config value

## Ownership / safety
- `ui/app.js` remains the sole owner of lock activation, config reads, credential comparison, error decision, and unlock state
- the new shell owner only observes `#lock-overlay` visibility and manages modal/focus/accessibility state
- no config schema, password persistence, IPC, WebView, account, broadcast, release/version, or subscription changes

## Design basis
- follows the W3C APG modal dialog pattern: underlying content inert, focus contained inside the dialog, visible focus, and logical focus return on close
- keeps Escape explicitly handled so the lock surface cannot be dismissed by modal keyboard convention

## Validation
- new `test/lock-screen-accessibility-contract.cjs` locks the ownership split and modal semantics
- new `e2e/specs/lock-screen-accessibility.e2e.cjs` verifies inert background, focus cycling, Escape resistance, accessible wrong-password state, and focus/interactivity restoration after unlock
- require exact-head repository tests, Electron E2E, and normal validation before merge


9529360-cpu:
Superseded by merged PR #460, which owns the same six-file lock-screen accessibility surface on the current architecture. This branch is now 27 commits behind live `master` and should not be rebased or merged as a second implementation line.

PR #482, opened 2026-09-14
fix: make authenticated proxies partition-scoped

### LINKED ISSUE #480: [P1] Make authenticated proxies partition-scoped and ready before WebView navigation
## Root cause

Geek exposes both global and per-account proxy username/password fields, but the current main-process proxy path is incomplete in two ways:

1. `src/main.cjs::proxyRulesFor()` embeds credentials directly into `proxyRules` as `user:password@host:port`. Electron's current `ProxyConfig.proxyRules` grammar accepts proxy scheme/host/port, not URL userinfo. Electron exposes authenticated proxy credentials through the `login` authentication challenge (`authInfo.isProxy`) instead.
2. `will-attach-webview` calls async `session.setProxy()` without any way to await it before the WebView attachment/navigation continues. Existing accounts therefore have no authoritative startup pre-application step, and newly created accounts can likewise reach their first navigation before the configured proxy operation has completed.

In a multi-account client this is not only a feature gap: proxy credentials and routing must stay bound to the account's persistent partition. A global credential callback or focus-based lookup could leak one account's proxy credentials to another account or to ordinary website Basic Auth.

Electron also documents that after changing a Session proxy, pooled sockets from the previous proxy may need to be closed so future requests do not reuse the previous route.

## Design

Add a small main-process proxy runtime owner and keep Account State / Config State as the only configuration authorities:

- build Electron `proxyRules` from scheme + host + port only; never place username/password in the rule string;
- install one `app` `login` handler after account/config state load and before any account WebView can be created;
- answer only `authInfo.isProxy === true` challenges;
- resolve the requesting `webContents.session.partition` back through `accountState.findByPartition()`;
- select the effective proxy from that account override or the current global config;
- require exact configured proxy host + port match before calling `event.preventDefault()` and `callback(username, password)`;
- never answer ordinary site Basic Auth, unknown/default sessions, mismatched proxy endpoints, or a repeated failed authentication attempt;
- pre-apply every existing account Session proxy before the main account UI can create WebViews;
- pre-apply a newly added account Session before returning it to the renderer;
- keep account/config update application post-durable-commit, but avoid reapplying/closing connections when the effective proxy configuration has not changed;
- when an existing Session's effective proxy/credentials do change, close its connection pool after `setProxy()` so future requests cannot keep using the previous route;
- retain a per-partition readiness result so a failed enabled-proxy apply can fail closed at WebView attachment rather than silently navigating direct.

## Scope

- new focused `src/proxy-runtime.cjs` owner;
- minimal `src/main.cjs` composition changes;
- focused executable proxy-runtime/integration contract;
- preserve existing encrypted-at-rest proxy password handling in Account State and Config State;
- no renderer redesign, account file/schema migration, WebView sandbox/navigation/permission relaxation, Subscription/Worker changes, package/version/release-marker/deployment changes;
- do not touch the open #457 write set.

## Validation

Executable coverage should prove at least:
1. generated proxy rules never contain username/password/userinfo and preserve HTTP/HTTPS/SOCKS routing semantics;
2. a proxy auth challenge for account A's exact partition + endpoint receives only A's credentials;
3. account B, an unknown/default Session, an endpoint mismatch, and non-proxy Basic Auth receive no credentials;
4. a repeated failed authentication attempt is not answered again;
5. global proxy credentials are used only for accounts without an account-specific override;
6. proxy application is deduplicated when the effective proxy fingerprint is unchanged;
7. a changed existing proxy calls `setProxy()` then closes that Session's pooled connections;
8. startup composition applies all existing account proxies before the subscription/main-window gate can create the account UI;
9. new-account composition applies its effective proxy before returning the account to the renderer;
10. an enabled proxy apply failure marks only that partition not ready, and WebView attach fails closed for that partition;
11. exact-head repository tests and aggregate Electron E2E pass before merge.

## References

Electron `ProxyConfig`: `proxyURL = [proxyScheme://]proxyHost[:proxyPort]`.
Electron `app` `login`: use `authInfo.isProxy`, `event.preventDefault()` and the credential callback for proxy authentication.
Electron `Session.setProxy()`: after proxy changes, `closeAllConnections()` may be needed to prevent pooled sockets using the previous proxy.

9529360-cpu:
Closes #480.

## What changed
- add a dedicated main-process proxy runtime for Electron Session proxy routing and proxy-auth challenges;
- emit proxy rules as protocol/host/port only, never credential-bearing userinfo;
- bind proxy credentials to the requesting account partition and exact proxy endpoint;
- preserve Account State / Config State as the only configuration authorities by decorating the factories consumed by legacy `main.cjs`;
- pre-apply existing account Sessions after both durable stores load and before account WebViews can attach;
- intercept legacy account `Session.setProxy()` calls and re-resolve the authoritative account/global proxy instead of trusting caller-built rules;
- close pooled connections only when the effective proxy fingerprint changes;
- fail closed per partition when an enabled proxy is invalid or cannot be applied;
- forget runtime readiness when an account is removed and restore native Session/store factories on dispose.

## Isolation and compatibility
- account A credentials cannot be used by account B, unknown/default Sessions, endpoint mismatches, ordinary site Basic Auth, or repeated failed auth attempts;
- non-account Sessions keep native Electron `setProxy()` behavior;
- existing WebView sandbox/navigation/permission boundaries are unchanged;
- encrypted-at-rest proxy passwords remain owned by Account State / Config State;
- no Subscription/Worker, website, package version, release marker, or deployment changes.

## Validation contract
`test/proxy-runtime-contract.cjs` covers protocol URI generation, credential omission, auth ownership, dedupe, connection-pool reset, per-account failure isolation, durable mutation ordering, startup pre-apply, legacy `setProxy` interception, fail-closed WebView attachment, account removal cleanup, dispose restoration, and `main-entry` startup gating.

The branch is merged with current `master` (`4ec2c4e`) before PR creation so CI validates the current website redesign and proxy runtime together.


9529360-cpu:
Superseded by merged #483 during this PR's final freshness barrier. #483 owns the account-partition proxy runtime and includes stronger same-endpoint credential-rotation handling (`clearAuthCache` + connection-pool invalidation), so this parallel implementation must not be merged as a second owner. I found one residual proxy-protocol grammar issue in the merged implementation and am extracting that as a minimal follow-up on current master rather than keeping overlapping runtime composition here.

PR #497, opened 2026-09-14
fix: reclaim committed account partitions after cleanup failures

### LINKED ISSUE #492: [P1] Reclaim committed account partitions after cleanup failures
## Root cause

Geek correctly treats Account State as the durable authority for account deletion, but the persistent Electron partition cleanup path has two holes after that durable delete commits:

1. `removeAccount()` only adds the partition directory to `pendingPartitionDeletions` when the later `fs.rm()` stage fails. If an earlier Session cleanup (`clearStorageData`, `clearCache`, `clearAuthCache`, `clearHostResolverCache`, or flush) throws, the function exits through the outer catch without registering the partition for exit cleanup.
2. startup orphan cleanup returns immediately when `snapshot.accounts.length === 0`. A successfully loaded empty Account State is authoritative (unrecoverable Account State load errors throw), so after deleting the final account any leftover `webview-page-*` directory can survive indefinitely across restarts.

Electron does not provide a documented API that deletes a persistent `persist:` partition directory itself; its session clear APIs clear data/cache but persistent directory reclamation is still an application lifecycle responsibility. Mature Electron desktop code also uses bounded Session cleanup plus filesystem deletion and startup orphan reclamation for crash/Windows-lock leftovers.

## Contract

- Account State remains the sole authority for whether an account/partition is live;
- after durable account deletion commits, any failure in Session cleanup must register that exact partition directory for the existing exit cleanup path before surfacing cleanup-pending semantics;
- startup orphan cleanup must run only after `await accountState.load()` succeeds, but a successfully loaded empty account list must be treated as authoritative and may reclaim account-shaped orphan directories;
- only `Partitions/webview-page-*` directories absent from the authoritative account partition set may be reclaimed;
- never delete unrelated `Partitions` entries or an active account partition;
- preserve current retry/fail-safe behavior and do not roll back a committed account deletion because child cleanup failed;
- no renderer, Worker, subscription, dependency, client-version, release-marker or deployment changes.

## Validation

Prove at minimum:

1. valid empty Account State + `webview-page-old` => orphan is selected for cleanup;
2. valid empty Account State does not select unrelated directories;
3. active partition is never selected while another orphan is;
4. a Session cleanup failure after durable account deletion registers the partition directory in `pendingPartitionDeletions`;
5. normal `fs.rm` failure still registers the same retry path;
6. orphan cleanup remains scheduled only after Account State load succeeds;
7. repository tests, Cloudflare validation and Electron E2E remain green.

References: Electron session clear APIs resolve when their specific data/cache operation completes, but there is no persistent-partition directory delete API; Electron issue #27142 tracks that gap. Comfy Desktop uses filesystem removal plus startup orphan reclamation for persistent partition cleanup.

9529360-cpu:
Closes #492.

## Root cause

Account deletion is durably committed before Electron session cleanup. The existing main-process path only registered `userData/Partitions/<partition>` for exit-time retry when directory removal itself failed. If an earlier Session cleanup step (`clearStorageData`, cache/auth/host-resolver cleanup, or `flushStorageData`) rejected after the account had already been removed from Account State, the persistent partition could remain on disk without entering the existing retry owner.

Startup orphan reclamation also returned early when the successfully loaded Account State contained zero accounts. That makes the final-account case sticky: `webview-page-*` directories can survive every restart even though an empty Account State is authoritative.

## Change

Keep Account State as the single authority and extend the existing cleanup owner only:

- resolve the exact committed account partition directory before any Session cleanup step;
- if Session cleanup fails after durable account removal, register that directory in the existing `pendingPartitionDeletions` retry set before surfacing cleanup-pending semantics;
- keep normal `fs.rm` retry behavior and register the same exact path on directory cleanup failure;
- clear a stale pending marker once the directory is successfully removed;
- allow startup orphan reclamation when Account State loaded successfully with zero accounts;
- retain the existing `collectOrphanPartitions` safety filter, so only inactive `webview-page-*` directories are reclaimable.

No renderer, Worker, subscription, dependency, package-version, release-marker, or deployment-path changes.

## Contract coverage

The existing partition helper contract already proves:

1. active partitions are retained;
2. unrelated directories are never selected;
3. an empty active-account list selects only `webview-page-*` orphan directories.

The integration contract now additionally locks:

4. orphan cleanup remains scheduled only after `await accountState.load()`;
5. a successfully loaded empty account list must not short-circuit reclaim;
6. the partition path is fixed before any Session cleanup call;
7. post-commit Session cleanup failures register the exact path for exit retry;
8. successful directory deletion clears stale pending state.

Exact-head repository tests, Cloudflare validation, Windows WhatsApp cold-start, and Linux Electron shell/E2E gates are required before merge.


9529360-cpu:
Validation note before merge: the implementation direction matches the durable Account State / persistent-partition ownership model, and the empty-account orphan fix is well scoped. One gap remains: the new integration contract proves post-commit Session-cleanup retry registration only by regex over `src/main.cjs`. Because that control-flow hole is the root cause of #492, please add an executable regression that injects a Session cleanup rejection after durable account removal and asserts the exact `Partitions/<removed partition>` path enters the existing pending/exit cleanup owner. Keep the existing fs.rm-failure case and empty-state orphan selection coverage. Static source matching can remain as a wiring guard, but it should not be the only proof of the failure path.

9529360-cpu:
Follow-up: I added executable failure-path coverage in commit `2792937e0419020559c306291494b990cde4742d`. The new contract executes the production `removeAccount()` owner with an injected `clearStorageData()` rejection and verifies the exact removed partition enters the exit-time retry set after durable account removal; it also executes the successful directory-removal path and proves a stale pending marker is cleared. Re-running exact-head CI now.

PR #499, opened 2026-09-14
test: execute partition cleanup failure recovery

9529360-cpu:
Follow-up evidence for #497 / #492.

The production fix is already in `master`; this PR adds the missing executable regression for the exact post-commit cleanup failure paths without changing runtime code.

It executes the real `removeAccount()` owner from `src/main.cjs` in a fault-injection harness and proves:

- a Session cleanup rejection after durable account removal registers the exact partition directory for exit-time retry and still publishes committed state from `finally`;
- repeated filesystem deletion failure keeps the existing bounded five-attempt retry behavior and the same pending retry path;
- successful directory deletion clears a stale pending marker;
- no renderer, Worker, dependency, version, release marker, deploy workflow, or production path changes.

Repository test, Worker validation, and Electron E2E must remain green before merge.


9529360-cpu:
Closing as a duplicate of #498. Both follow up #497/#492 with the same executable `removeAccount()` fault-injection coverage, including post-commit Session cleanup failure, bounded filesystem deletion retry exhaustion, and stale pending-marker cleanup. #498 was opened first and already has repository tests + Worker validation green, with Electron E2E in progress. Keep one root cause / one PR.

PR #533, opened 2026-09-14
fix: cancel translated sends after chat switches

### LINKED ISSUE #527: [P0] Cancel Telegram/LINE outgoing translation when chat changes before send
## User-visible risk

Telegram and LINE outgoing translation can complete after the user has navigated to another chat. The current send adapters capture the original `chatId`, await `window.__geekTranslationRequest(...)`, then mutate the currently mounted composer and trigger a programmatic send without re-validating that the current chat is still the original chat.

If the platform reuses or replaces the composer while translation is pending, the translated text can be written/submitted in a different conversation. This is a message-integrity / wrong-recipient risk, not only a latency issue.

WhatsApp already has the required precedent in `ui/whatsapp-translation-hook-recovery.js`: after translation resolves, it verifies the active chat still matches the attempt chat and cancels the send if the chat changed.

## Required behavior

1. Telegram and LINE outgoing translation must capture the initiating chat identity before starting translation.
2. After the translation Promise resolves and before any composer mutation/native input request, verify the active chat is still the initiating chat.
3. Verify again immediately before the final synthetic click/Enter, because chat state can change while native composer fill is in progress.
4. If the chat changed, cancel the operation, never mutate/send in the new chat, release the send lock, and show an actionable cancellation notice.
5. Never fall through to raw source text when outgoing translation is enabled.
6. Repeated trusted Enter/click behavior remains blocked while translation is pending.
7. Preserve existing Telegram/LINE platform transport and WebView/security boundaries.
8. Add executable regression coverage with a deferred translation request: start send in chat A, switch to chat B before resolution, resolve translation, and prove no native composer write / synthetic send occurs in chat B.
9. Cover the second race window by switching after translation but during/after composer fill before final submit.
10. No package version or release-marker changes.

## Validation

- focused Telegram/LINE outgoing translation chat-generation contract;
- existing translation double-enter / platform / send-safety contracts;
- full repository tests + Electron E2E;
- authenticated Telegram and LINE client validation remains the final production evidence boundary.

## Mature precedent

This follows the repository’s existing WhatsApp chat-switch guard and the standard async-generation rule used by mature UI systems: results created for generation/chat A may commit only if generation/chat A is still current.

9529360-cpu:
Closes #527.

## Root cause

Telegram and LINE outgoing translation captured the initiating chat, awaited asynchronous translation, and then could mutate/submit whatever composer was current without re-validating the conversation generation. Telegram also crosses the Electron native-input bridge during composer fill, creating a second race window.

## Change

- revalidate Telegram and LINE chat + composer identity after translation and again immediately before final submit;
- reject stale/replaced composer generations instead of touching the new chat;
- carry Telegram's expected chat identity as a **request-scoped, versioned, token-bound native-input envelope**;
- decode and validate that lease in the authoritative main-process WebView input owner immediately before `webContents.insertText`;
- keep raw/native-input callers backward compatible: requests without the envelope retain the existing path and cannot inherit any prior translation lease;
- remove the previously proposed page-global expected-chat state so translation cannot cross-couple Broadcast or another native-input user;
- reject malformed or wrong-token lease envelopes fail-closed before composer mutation;
- keep translation-enabled sends fail-closed and preserve repeated-send locking;
- strengthen the double-enter contract to assert the actual invariant (one programmatic submit, with a final chat-context guard) instead of one exact source expression.

No Translation Runtime scheduling/quota/provider behavior, Broadcast transport semantics, account/session ownership, Worker deployment, package version, or release marker is changed.

This PR is containment for #527 only. The architectural fix remains #542: outgoing send becomes the transaction owner and translation becomes a cancellable transform.

## Validation — exact head `291b507697b6f7ab108cdb624c271625ef75f616`

Executable coverage now proves:

1. Telegram: switch chat before translation resolves -> no native composer mutation and no send;
2. Telegram: switch during native composer fill -> no final synthetic send;
3. LINE: switch before translation resolves -> no composer mutation and no send;
4. LINE: switch after fill but before submit -> no final submit;
5. native-input owner: stale request-scoped chat lease is rejected before `insertText`;
6. wrong-token / malformed lease envelopes are rejected before composer mutation;
7. a later plain raw/native-input request still succeeds, proving stale translation state cannot leak into Broadcast/other native-input users;
8. repository test workflow: success;
9. Cloudflare validation/dry-runs: success;
10. Electron E2E: success, including Windows WhatsApp cold-start and Linux shell smoke.

Authenticated Telegram/LINE client validation remains the final production evidence boundary; this PR does not claim that external evidence.


9529360-cpu:
Architecture review after this PR was opened found a deeper root cause. Do **not** treat this PR as the main translation/send fix or merge it merely because focused CI becomes green.

#542 now records the replacement architecture: outgoing message send becomes the first-class transaction owner; translation becomes a cancellable transform stage; background receive/history translation moves to a separate QoS plane; the WebView bridge becomes terminal-response transport/security only.

This PR is therefore containment for #527 only (wrong-chat / stale composer risk). Its first implementation also used a page-global expected-chat lease for Telegram native input, which can interfere with concurrent non-composer native-input users such as Broadcast. That shared lease must not ship. If #527 containment is still needed before the architecture migration reaches TG/LINE, rework it to a per-request binding and validate Broadcast isolation on the exact head.

No merge claim until that containment is isolated and the architecture sequencing in #542 is respected.

9529360-cpu:
Blocking architecture review: the current guard does not actually bind `expectedChatId` to the native-input IPC request.

`ui/translation-adapters.js` now stores `expectedChatId` in `__geekNativeInputPending`, and `ui/app.js` already retrieves that pending record, but the host still calls `window.api.webviewInput.insertText(account.id, guestId, text, suppliedToken)` without the expected chat identity. `src/preload.cjs` and `src/webview-ipc.cjs` likewise expose only `(accountId, guestId, text, token)`. The main-process `FOCUSED_COMPOSER_SCRIPT` therefore reads `window.__geekNativeInputExpectedChatId`, a shared mutable page-global value, rather than the lease belonging to the exact request being authorized.

That violates the #542 invariant that native commit must be bound to exact request/account/WebView/conversation identity and specifically that expected-chat binding must be per request, not shared page-global state. It also leaves failure modes where a later request or timeout can overwrite/clear the global before the earlier main-process check; if it becomes empty the current script skips the chat comparison entirely.

Required correction before this PR is mergeable:

- carry `expectedChatId` from the pending record through `app.js -> preload webviewInput.insertText -> webview:insert-text` as an explicit request argument;
- validate/bound it at the trusted host/main boundary;
- execute the composer/chat check against that explicit argument, not `window.__geekNativeInputExpectedChatId`;
- remove the shared page-global expected-chat owner (the pending map can remain per-request);
- add regression coverage proving request A's expected chat cannot be changed/bypassed by request B/global mutation/timeout before A reaches `insertText`.

The existing chat-switch tests prove the happy stale-chat rejection but do not prove per-request lease ownership, so green CI is not sufficient for this version.

9529360-cpu:
Additional exact-head evidence after refreshing onto current translation foundation: repository `test` is red in the existing `translation-double-enter-contract.cjs` (`Telegram 必须保留译文的程序化提交`; expected `(button || sendButton())?.click()`). So this head is blocked both by the per-request native-input lease issue above and by a concrete compatibility regression in the established Telegram send contract. Do not treat Electron/other checks as sufficient until both are corrected.

PR #557, opened 2026-09-14
test: lock translation remote singleflight contract

### LINKED ISSUE #529: [P1] Register translation single-flight before async auth so identical requests cannot double-call or double-charge
## User-visible / billing impact

Concurrent identical translation requests can escape Geek's intended in-flight deduplication, consume extra remote capacity/rate-limit budget, and be billed as distinct translation requests. This can indirectly make ordinary WhatsApp outgoing translation fail while broadcast remains healthy.

## Confirmed race

`src/translation-runtime.cjs` computes `inflightKey = partition + cacheKey` and checks:

```js
if (state.inflight.has(inflightKey)) return state.inflight.get(inflightKey);
```

But the authoritative in-flight Promise is not installed until after several async boundaries:

1. local quota lookup;
2. `await getSubscriptionStore().getTranslationToken()`;
3. request-id allocation / remote queue construction;
4. only then `state.inflight.set(inflightKey, request)`.

Two identical requests entering together can both observe no in-flight owner, await independently, then each create a different UUID / `X-Request-ID` and enqueue its own `/v1/translate` call. Server billing deduplicates by request ID, so two distinct IDs are two legitimate reservations/charges.

`getTranslationToken()` itself only caches an already-completed token; it has no in-flight single-flight Promise. A cold burst can therefore duplicate the token POST as well.

The existing `translation-cache-singleflight-contract.cjs` proves single-flight for the persisted cache-file *load*, not remote translation execution, so its name currently overstates the protected mechanism.

## Required behavior

1. Establish the authoritative same-key in-flight owner before any async work that followers should share.
2. Same partition + same translation cache key + same effective refresh semantics must share one remote translation outcome and one billable request where dedupe is valid.
3. Followers must not each fetch a short-lived translation token while one same-generation acquisition is already pending.
4. Preserve `refresh: true` semantics deliberately: decide whether concurrent refreshes coalesce, and encode that contract explicitly rather than accidentally.
5. Account deletion/session-generation changes must reject all followers consistently and must not leave a stale Promise in the map.
6. One failed leader must clean up the key so a later legitimate retry can proceed.
7. Keep cross-partition requests isolated; identical text in two accounts must never share authorization/cache/billing ownership.
8. Preserve server request-id idempotency as the outside-process safety net; do not replace it with client-only dedupe.
9. Add deterministic deferred tests proving 20+ simultaneous identical requests produce exactly one token acquisition where applicable, one gateway request and one request ID, while two accounts remain independent.
10. No package version / release-marker changes.

## Validation

- new remote single-flight contract (not only cache-load single-flight);
- subscription token acquisition single-flight contract;
- account deletion/session-generation contracts;
- quota/billing Worker contract;
- full repository tests + Electron E2E.

9529360-cpu:
Closes #529 as a regression-coverage follow-up to merged #543/#544.

## Live-code finding

The Issue was correct when filed, but live `master` no longer has the described race. #543 moved cache/quota/token/remote work inside one leader Promise and registers that Promise in `state.inflight` synchronously before the handler yields back to the event loop. #544 separately singleflights cold translation-token acquisition by session generation.

So this PR intentionally makes **no production-code change**. It adds the missing billing/concurrency contract so a future scheduler/SendIntent refactor cannot regress the repaired behavior.

## Regression coverage

`test/translation-remote-singleflight-contract.cjs` proves:

1. 25 simultaneous identical cold requests in one partition produce exactly one quota preflight, one token acquisition, one request ID and one gateway call;
2. all followers receive the leader request ID/result;
3. a failed leader is shared by followers but releases authority so a later retry performs fresh work with a new request ID;
4. identical text in two account partitions remains isolated and produces distinct request IDs/gateway calls;
5. concurrent `refresh:true` calls deliberately remain independent, documenting the current refresh semantic rather than accidentally coalescing manual refresh.

## Precedent

This matches the established singleflight model: duplicate in-flight work for the same key shares one leader result, and the key is forgotten after completion rather than becoming a cache authority.

The final diff is exactly one new test file. No Translation Runtime/Worker, WebView, broadcast, package version or release-marker changes. Current head `ade1f0aea0542ccb962b8368c6342d8b304f42f4` is one commit ahead of live master `adf86df14fba2be73d2728b36f21327872ab6f18`, behind by 0, and remains disjoint from #533/#551/#554/#556.

## Exact-head validation

On `ade1f0aea0542ccb962b8368c6342d8b304f42f4`:

- repository `test`: success;
- `electron-e2e`: success;
- `cloudflare-worker-validation`: success.

This is regression evidence for repository behavior, not authenticated real-client evidence.


9529360-cpu:
Freshness update: this regression-only PR was rebuilt as one test-file commit on master `a9e5cd82222d91f74b226fbaa9fa2ae92f56e253`; current head is `494e92524e5dbc5b92be71371d5e9a7e01fc1e95`. Older exact-head SHA claims in the body are stale. Current-head repository tests and Worker validation are green; Electron E2E has not yet reached terminal success, so do not merge until the exact current head is green.

Issue #526, opened 2026-09-14
[P1] Add a privacy-safe real-client WhatsApp composer translation smoke matrix


9529360-cpu:
## Problem

Geek now has multiple independent regressions where ordinary WhatsApp composer sends fail/block while direct WPP broadcast remains healthy (#489, #493, #516, plus current #518–#524 findings). CI contracts validate mechanisms, but the product evidence boundary still does not exercise the real authenticated composer + translation path.

`docs/windows-real-client-runner.md` correctly separates real Windows evidence from CI. However `scripts/windows-real-client-evidence.ps1` only captures structural state (installed version/process, account/config file metadata, partition count, legacy LINE token-store presence). It intentionally does not prove authenticated WA/TG/LINE messaging behavior.

As a result, exact-head tests/Electron smoke can all be green while the user-critical combination `real WhatsApp page + current Meta module graph + persistent upgraded profile + translation bridge + subscription auth + visible-message load + composer send` is unverified.

## Required evidence matrix

Create an owner-run, privacy-safe manual/assisted real-client checklist or local harness for a dedicated test conversation/account. Do not turn a real profile into a GitHub self-hosted runner and do not upload account/chat data.

At minimum validate on the candidate build/revision:

1. Translation OFF: ordinary WhatsApp Enter and Send-button are pure native pass-through.
2. Translation ON: one ordinary send translates and submits exactly once; source draft is recoverable on translation failure.
3. Busy chat: enable automatic received-message translation, open a chat with enough visible messages to create meaningful concurrency, then immediately send; composer must not wait for background translation or hit a 35s orphan timeout.
4. Reload/cold start: repeat ordinary send after WebView reload and app restart to exercise bridge marker and WA-JS module-generation recovery.
5. User readiness states: valid quota, zero quota, logged-out/expired subscription state; errors must be actionable and must never masquerade as native WhatsApp transport failure.
6. Upgraded profile: use a copy/owner-controlled profile with historical translation settings to prove stale provider/route values are normalized.
7. Addressing: test a normal direct chat and an LID-backed/current WhatsApp direct chat where available.
8. Broadcast text remains a positive control, but broadcast success must not be used as proof that the composer path passed.

## Privacy-safe diagnostics during the smoke

Record only categorical/non-sensitive state needed to classify the failure, for example:

- app version/revision and WA-JS package version;
- bridge marker present/missing;
- send-hook generation ready/waiting;
[...]

PR #563, opened 2026-09-14
test: lock translation queue wait into caller deadline

### LINKED ISSUE #519: [P1] Bound translation queue wait inside one end-to-end request deadline
## User-visible failure

Ordinary WhatsApp composer translation can time out/fail even while the underlying account and WPP send transport are healthy. Broadcast remains unaffected because it does not wait on the ordinary composer translation request.

This is independent of #516/#517 and the WebView bridge saturation bug tracked separately in #518.

## Confirmed mechanism

The page-side translation Promise has a fixed 35s timeout and deletes its pending request when that timer fires.

The main-process Translation Runtime owns one global `remoteQueue` / `remoteActive` pool with `TRANSLATION_REMOTE_LIMIT = 20`. A translation request is admitted to that queue before the remote translation function runs.

Inside the queued task, the runtime currently creates:

```js
const deadline = Date.now() + 30000;
```

So the 30s network/failover deadline starts only **after queue wait has finished**. Queue time is unbounded by that deadline. Translation-token acquisition also happens before queue admission and has its own request budget.

Therefore end-to-end time can be:

`subscription token acquisition + global queue wait + up to 30s gateway attempts`

while the WebView caller is willing to wait only 35s total. Under load, the page can time out and delete the request while main-process work later continues, potentially completing translation that can no longer be delivered to the caller.

Because the remote pool is global, background/history translation from one account can also delay an interactive outgoing send from another account.

## Required behavior

1. Establish one absolute end-to-end deadline when a translation request is accepted by Translation Runtime.
2. Token acquisition, queue wait, failover and gateway fetch must consume that same budget rather than starting independent full budgets later.
3. A queued request whose deadline has already expired must reject before starting a gateway request.
4. AbortController timeout must use the remaining end-to-end budget.
5. The runtime must finish/reject with enough margin before the page-side caller timeout; do not allow successful main-process work to routinely arrive after the page has discarded the request.
6. Preserve bounded global concurrency, cache/inflight dedupe, account-deletion cancellation, gateway failover and quota/auth rules.
7. Do not change broadcast transport or fail open to raw source text.
8. Add a deterministic queue-saturation contract with controlled deferred work proving a later request's queue wait consumes its original deadline and no post-deadline remote fetch starts.
9. Add cross-account coverage so one account cannot create unbounded invisible latency for another account's request.
10. No package version / release-marker changes.

## Validation

- focused Translation Runtime queue/deadline contract;
- existing failover, cache, account-deletion and quota contracts;
- full repository tests + Electron E2E;
- real logged-in WhatsApp send under concurrent translation load remains the final production evidence boundary.

9529360-cpu:
Closes #519 as regression coverage for deadline behavior already present in live Translation Runtime.

## Live-code finding

The Issue was valid when filed, but current Runtime now creates one caller deadline before cache/quota/token/remote work, carries that absolute deadline into `enqueueRemote()`, arms a queue timer using the remaining budget, removes an expired item from the queue, and re-checks remaining budget before starting remote work.

This matches the standard end-to-end deadline model: time already spent waiting must be deducted, and work should not begin after the caller has stopped caring.

## Change

No production code changes. Add `test/translation-runtime-queue-deadline-contract.cjs`.

The contract proves:

1. all `TRANSLATION_REMOTE_LIMIT` slots are occupied by distinct account-A requests;
2. an account-B request enters with a short absolute deadline while capacity is unavailable;
3. the account-B request rejects with `TRANSLATION_DEADLINE_EXCEEDED` while still queued;
4. gateway invocation count remains exactly the remote limit, proving expired queued work never starts network/billable work;
5. after all active account-A requests finish and capacity returns, the expired account-B request is not resurrected.

## Freshness / exact-head validation

Current head `1f82d4668096163fb3ae1cbcebae34344f6bde84` is refreshed on live master `4a76de2f12a914936bf582b4808d9bc043e2fde7`, behind by 0. Final diff is one new test file only. No Translation Runtime/Worker, WebView, broadcast, package version or release-marker changes.

On `1f82d4668096163fb3ae1cbcebae34344f6bde84`:

- repository `test`: success;
- `electron-e2e`: success;
- `cloudflare-worker-validation`: success.

Real authenticated dense-chat send latency remains a separate product-evidence boundary (#526/#560); this PR locks repository behavior only.


9529360-cpu:
Superseded by #569, which reapplies the same final test-only contract onto current master `2d1deb460e25dcf7ae11a749298e6e429f284578` so exact-head CI covers the current combined tree. Closing this stale candidate to keep one active authority.

Issue #542, opened 2026-09-14
[Architecture] Make outgoing message send the transaction owner; demote translation to a cancellable transform


9529360-cpu:
## Decision

The current failure cluster is not a collection of independent translation bugs. The original 2026-08-15 architecture made the shared translation core/request queue the owner of both background message translation and send-critical outgoing translation. Later hardening changed transports, security and module ownership, but preserved that premise.

That premise is now the root architectural defect.

**Outgoing user send must become the first-class transaction owner. Translation must be only one cancellable transform stage inside that transaction.** Background receive/history translation must be a separate low-priority work plane. The WebView bridge must be transport/security only, not a second scheduler.

## Historical evidence

`docs/多平台翻译架构研究.md` explicitly adopted the HaloGPT model where send and received messages use one common translation core, one request queue and one concurrency bound. Telegram/LINE then copied the original-product send-interception model. WhatsApp later added a global Promise send queue when rapid sends reordered/merged.

The repository has since accumulated compensating layers around that decision:

- page pending Promise maps + 35s timers;
- bridge inflight admission;
- renderer WebView dispatch;
- Subscription quota/token preflight;
- TranslationRuntime global queue/concurrency;
- desktop gateway deadline/failover;
- Worker provider retry/timeouts;
- platform-specific send locks / synthetic submits;
- WhatsApp composer-intent reconstruction from chat/text/TTL around a low-level monkey patch.

FH-04 / PR #349 correctly extracted TranslationRuntime ownership from `main.cjs`, but intentionally preserved behavior. It did not create a message-send transaction model.

By contrast, Broadcast already uses the healthier architecture: a first-class Job freezes account/partition/platform/WebView/targets/message/attachments ownership, has an explicit state machine, uses `GeekPlatformTransports`, and revalidates chat/composer through `GeekBroadcastSafety.authorizeSend` before commit.

`ui/platform-adapter-contract.js` also states the intended rule directly: **translation and broadcast share platform capabilities, not business state**. Ordinary translated send currently violates that rule by owning send lifecycle inside platform-specific translation hooks.

### Interactive send is not an entity

There is no immutable SendIntent carrying:

- transaction ID;
- account / partition;
- exact WebView generation;
- conversation identity;
- composer generation/snapshot;
- original text transformation policy;
- absolute deadline;
- cancellation signal;
- explicit lifecycle state.

[...]

Issue #534, opened 2026-09-14
[P1] Close translation timeout-after-commit so a lost response cannot charge quota without a recoverable result


9529360-cpu:
## Value/message integrity failure

A translation can successfully complete and consume quota at the gateway, while the desktop loses/times out on the response. Retrying the same logical request then returns `duplicate_request` rather than the translated result. The user can end up with no sent message and less quota.

## Confirmed mechanism

Desktop Translation Runtime creates one UUID / `X-Request-ID` and reuses it across gateway endpoint attempts.

The gateway uses `translation_usage.request_id` as the idempotency/billing key:

1. reserve source-character quota;
2. call provider;
3. deduct target-character quota and mark `translation_usage.status = 'complete'`;
4. return the translated text.

The schema deliberately stores billing metadata only and **does not store chat source/translated text**.

If step 3 commits but the HTTP response in step 4 is lost/aborted, the client cannot know the outcome. A retry with the same request ID finds the existing usage record and returns `409 duplicate_request`. The translated result is not recoverable from the idempotency record.

This is classic timeout-after-commit ambiguity: request-id uniqueness prevents double charge but does not make the operation replayable.

## Required behavior

1. Define an explicit outcome-recovery contract for `X-Request-ID`: after any unknown transport outcome, the client must be able to determine whether the request is `not_started`, `reserved/in_progress`, `completed`, or `refunded/failed`.
2. A completed request must have a privacy-compatible way to deliver/recover a usable translation result, or the billing contract must compensate safely; "409 duplicate" is not a complete idempotency contract.
3. Preserve the repository privacy boundary: do not casually persist plaintext chat source/translation in D1/Issues/logs. If temporary encrypted replay storage is chosen, define key ownership, TTL, deletion and threat model. If a no-content recovery protocol is chosen, prove it cannot be abused for free translations/refunds.
4. Bind request ID to a stable hash of effective request semantics (user/account, source text, source/target language, provider/route/version) so the same ID cannot be replayed with different content.
5. Client retry/failover must reuse the same operation identity and classify `in_progress` separately from terminal duplicate/conflict.
6. Provider/gateway deadline cancellation must not silently convert a committed operation into an unrecoverable client failure.
7. Quota reservation/refund/finalization must remain atomic and exactly-once in value terms under response loss, worker restart and duplicate delivery.
[...]

PR #564, opened 2026-09-14
fix: bind translation work to subscription authorization lease

### LINKED ISSUE #540: [P1] Bind queued translation work to the subscription session generation that authorized it
## Cross-session value-integrity risk

Translation work can outlive the Geek subscription/login session that authorized it. A queued request may continue using an old short-lived translation JWT after the user logs out or switches to another subscription account, charging the old user's quota after that session is no longer current.

## Confirmed lifecycle gap

`subscription.cjs::getTranslationToken()` correctly captures `sessionGeneration` while obtaining/caching the 5-minute translation token and rejects if the session changes during acquisition.

Translation Runtime then stores the returned token in local request state **before** remote-queue admission:

```js
const remoteAuthorization = await getSubscriptionStore().getTranslationToken();
...
const request = enqueueRemote(partition, async () => {
  // later fetch using captured remoteAuthorization
});
```

The queued/active translation is bound to the local messaging account partition but not to the subscription-store session generation. Logout/login clears the store token/cache and increments its generation, but Translation Runtime is not notified and does not cancel/revalidate already-authorized queued work.

The short-lived JWT remains cryptographically valid until expiry, so old queued work can still reach the Worker as the previous subscription user.

## Required behavior

1. Every remote translation must carry/capture the subscription session generation (or an opaque authorization lease identity) that issued its translation token.
2. Before queue execution and before committing/delivering a result, verify that authorization lease is still current; logout/account switch must cancel or reject stale queued work.
3. Active provider/gateway work should be aborted on a subscription-session invalidation where practical, coordinated with #534 cancellation/outcome reconciliation.
4. A late old-session response must never mutate the new session's local quota projection (#538), token cache, UI readiness, or any value state.
5. Do not bind this to UI focus/active messaging account; the authority is the subscription session, while messaging partition identity remains independently enforced.
6. Preserve legitimate same-session account switching between WhatsApp/TG/LINE accounts if the product intentionally shares one Geek quota.
7. Add deterministic ABA coverage: session A request queues -> logout -> login B -> old A token/request resolves; prove no B state mutation and define whether A operation is canceled/reconciled.
8. Add same-user re-login generation coverage so equality of user/account fields does not let a stale request cross a generation change.
9. No package/release-marker changes.

## Validation

- subscription session-generation contract;
- Translation Runtime queued/active authorization lease tests;
- logout/login ABA tests;
- #534/#538 outcome/quota reconciliation interaction;
- full repository tests + Electron E2E.

9529360-cpu:
Closes #540. Part of #542.

## Root cause

TranslationRuntime currently captures a short-lived translation JWT before remote-queue admission and then treats that bare string as valid until the request finishes. Logout/account switch correctly clears the Subscription store and increments its session generation, but already-authorized queued/active work can keep using the old JWT until expiry. That can consume the old subscription user's quota after a new login/session becomes current.

## Design

This PR treats authorization as a call-lifecycle lease rather than a one-time token lookup:

- Subscription remains the sole session authority;
- `getTranslationAuthorization()` returns a frozen `{ token, generation, signal }` lease tied to the current subscription generation;
- the generation owns an AbortSignal; only after logout/account-switch tombstone persistence succeeds does Subscription increment generation and abort the old signal;
- TranslationRuntime binds remote queue admission to that signal, removes queued stale-session work before network execution, and links active gateway AbortControllers to session invalidation;
- Runtime revalidates the lease before network work and before a remote result may update cache or return to the caller;
- `SUBSCRIPTION_SESSION_CHANGED` is caller lifecycle/auth state, never gateway-health evidence;
- legacy `getTranslationToken()` remains available for existing callers/tests, while the real production store uses the stronger lease API.

This follows the established call-context model used by mature RPC systems: authorization, deadline and cancellation are properties of the operation lifecycle, not independent snapshots.

## Safety / value boundary

- Same subscription session may still serve multiple WA/TG/LINE messaging partitions; messaging-account ownership remains independent.
- Same-user relogin still creates a new generation, so ABA identity equality cannot revive old work.
- A transport that ignores AbortSignal is still blocked by post-response lease revalidation before local cache/result commit.
- This does not claim to solve timeout-after-server-commit billing ambiguity; #534 owns outcome recovery.
- No Worker, D1 schema, package version, release marker, broadcast, WebView bridge or platform-send changes.

## Regression coverage

`test/translation-subscription-authorization-lease-contract.cjs` covers:

1. real Subscription lease creation and durable logout invalidation;
2. 20 active + 1 queued remote translations, then session invalidation: queued work never starts, active calls abort, gateway health is unchanged;
3. a fake transport that deliberately ignores AbortSignal and returns late: stale result is rejected and never reaches cache persistence.

Exact-head CI is required before merge. Authenticated real-client login/account-switch behavior remains a separate evidence boundary.


9529360-cpu:
Superseded by #577, which reapplies the authorization lease onto current master after #571/#576 landed. #577 explicitly preserves route-aware endpoint selection, HTTP 413 request taxonomy, and half-open success/failure/inconclusive lifecycle; exact-head CI will cover the combined tree.

Issue #522, opened 2026-09-14
[P1] Revalidate WhatsApp runtime on current stable WA-JS instead of locked 4.3.0


9529360-cpu:
## Why this matters

Geek currently declares `@wppconnect/wa-js: ^4.3.0`, but `package-lock.json` resolves the production dependency to exactly **4.3.0**.

Since 4.3.0, upstream stable WA-JS releases have shipped multiple compatibility fixes for WhatsApp Web 2.3000, including:

- 4.4.0: restore module bindings broken on newer WhatsApp builds, recover stale negative module lookups, and LID send fixes;
- 4.4.3: compatibility with WhatsApp Web >= 2.3000.1044096409;
- 4.5.0: wait for the Meta module graph to settle before `loader.injected`, fixing cold-load `isReady` / `isFullReady` / module-not-found failures; additional LID chat fixes;
- 4.6.0: further chat/LID/runtime compatibility fixes.

The upstream WPPConnect project itself moved to WA-JS 4.6.0 in its 2026-08-15 release.

This is directly relevant to Geek because ordinary composer translation recovery now intentionally follows WA-JS module metadata (#516/#517), while broadcast also depends on WPP APIs. A healthy broadcast call does not prove every WA-JS module binding / loader path used by the composer hook is compatible.

## Required work

1. Treat this as a compatibility migration, not a blind dependency bump.
2. Compare the Geek-used WA-JS surface against current stable 4.6.0 release notes/contracts, especially:
   - `WPP.loader` / module metadata used by WhatsApp translation hook recovery;
   - `WPP.chat.sendTextMessage` and send-file/media paths used by broadcast;
   - active/current chat lookup;
   - LID / phone-number addressing;
   - group participant / mention paths used by group tools and broadcast;
   - injection readiness and cold-start behavior.
3. Update both manifest and lockfile to a deliberately tested stable version; do not leave runtime behavior dependent on an unreviewed floating minor version.
4. Preserve `resources/waplus-wpp.js` compatibility or explicitly adapt its boundary if necessary.
5. Add/refresh executable contracts around the exact WA-JS APIs Geek consumes rather than only checking that the bundle injects.
6. Run Windows WhatsApp cold-start, authenticated ordinary composer send, broadcast text/media, LID/direct chat, group send/mentions, reload and restart/session persistence checks.
7. Keep package/release marker unchanged; dependency maintenance is not a formal client release by itself.

## Evidence boundary

[...]

Issue #516, opened 2026-09-14
[P1] Resolve WhatsApp composer translation send owner through WA-JS instead of one private module name


9529360-cpu:
## User-visible failure

Ordinary WhatsApp composer messages can be blocked/fail while the same account still sends through Geek broadcast. The current 1.2.25 tree already contains #491/#494 recovery/failure-boundary work, but both `ui/app.js::syncTranslationCfgToWebview()` and `ui/whatsapp-translation-hook-recovery.js` still resolve the native composer send owner only through `window.require('WAWebSendTextMsgChatAction')`.

If that private Meta module name is unavailable, registered later, or changes generation, the translation safety guard correctly fails closed and ordinary composer sends stop, while broadcast remains healthy because it uses the WPP/platform transport instead of this composer wrapper.

## Current upstream precedent

Current `@wppconnect/wa-js` does not bind `sendTextMsgToChat` by a fixed private module name. Its `exportModule` searches modules by capability (`m => m.sendTextMsgToChat`), records the resolved module id in `WPP.whatsapp._moduleIdMap`, and exposes the active loader as `WPP.loader.moduleRequire`. Geek already injects WA-JS into WhatsApp WebViews.

## Required change

1. Add one WhatsApp send-module resolver owned by the existing recovery path.
2. Prefer the module id already resolved by WA-JS for `WPP.whatsapp.functions.sendTextMsgToChat`, then load that exact live module through `WPP.loader.moduleRequire`.
3. Keep `WAWebSendTextMsgChatAction` only as a backward-compatible fallback.
4. Make the existing app wrapper/raw-send guard and lifecycle recovery use the same resolver; do not create a parallel send authority.
5. Preserve fail-closed behavior when translation is requested but no compatible live send owner exists; never leak raw source as a fallback.
6. Preserve #494 translation failure isolation, compose restoration, chat-switch protection, repeated-send blocking, serialized ordering, and outgoing-original mapping.
7. Add dynamic coverage where `window.require('WAWebSendTextMsgChatAction')` is missing but WA-JS module-id resolution succeeds; ordinary translated sending must remain healthy.
8. Translation-disabled chats remain native pass-through.
9. No broadcast transport, Worker/quota/provider, account/session security, dependency/version marker, or release changes.

## Validation

- focused WhatsApp translation hook recovery contract;
- translation send safety contract;
- repository tests and Electron E2E via PR CI;
- real logged-in WhatsApp composer validation remains the final production evidence boundary.

Issue #524, opened 2026-09-14
[P1] Fix the broken translation bridge fallback when the preload marker is absent


9529360-cpu:
## User-visible failure

Ordinary WhatsApp outgoing translation can fail immediately while broadcast remains healthy if the guest preload bridge marker is temporarily or permanently unavailable.

## Confirmed bug

The page-side translation request code in both the WhatsApp injection (`ui/app.js`) and shared adapters (`ui/translation-adapters.js`) uses:

```js
if (
  document.documentElement.getAttribute('data-geek-bridge') === '1' ||
  document.getAttribute('data-geek-bridge') === '1'
) {
  // postMessage -> preload sendToHost
} else {
  // console-message compatibility fallback
}
```

Standard `Document` does not implement `getAttribute()`. The second operand is normally hidden by short-circuiting while `document.documentElement` carries the marker. But if the marker is absent, the code throws before reaching the intended console-message fallback.

That makes the fallback path self-defeating exactly when it is needed: preload readiness/load-order/navigation/packaging drift turns into an immediate rejected translation request, which blocks translation-enabled ordinary sends. Broadcast does not use this page translation bridge.

## Required behavior

1. Bridge readiness detection must only call valid DOM APIs and must tolerate `documentElement` not yet existing.
2. If the authenticated preload bridge marker is present, use the `postMessage`/`sendToHost` path.
3. If the marker is genuinely absent but the existing console compatibility path is still an intentional supported fallback, it must execute without throwing and remain token/request-id validated by the host.
4. If the console fallback is no longer considered an acceptable production path, remove it deliberately and return an immediate explicit terminal “bridge unavailable” error instead of a JavaScript TypeError/35s timeout.
5. Apply one shared readiness helper/contract to WhatsApp, Telegram and native-input requests instead of duplicating the same unsafe condition.
6. Preserve `contextIsolation=true`, token binding, same-origin postMessage checks and fail-closed raw-send semantics.
7. Add executable tests for marker present, marker absent, missing `documentElement`, and fallback/terminal-error behavior.
8. Broadcast transport remains unchanged.
9. No package version / release-marker changes.

## Validation

- bridge preload contract;
- WebView bridge security contract;
- WhatsApp outgoing-send translation contract;
- Telegram/LINE adapter bridge contracts where applicable;
- full repository tests + Electron E2E;
- real Windows cold-start/reload with bridge marker observation remains the final production evidence boundary.

9529360-cpu:
[...]

PR #573, opened 2026-09-14
fix: migrate WhatsApp runtime to WA-JS 4.6.0

9529360-cpu:
## Summary
- pin `@wppconnect/wa-js` to exact stable `4.6.0` without changing the formal client release marker;
- align injection lifecycle with official WA-JS readiness and keep `window.WPP` as the primary runtime, with `window.WAPLUS_WPP` only as a capability fallback;
- prefer WA-JS 4.6 `phoneNumber` for LID -> phone mapping while retaining legacy `pn` compatibility;
- keep GT/TG/LINE/common Translation Runtime as the shared translation authority; WhatsApp no longer owns a parallel translation implementation;
- reduce ordinary WhatsApp private translated-send behavior to one thin native-send adapter;
- keep translation/config/identity uncertainty fail-closed so raw source cannot leak when outgoing translation is enabled;
- preserve actionable Translation Runtime failures through Electron/contextBridge without exposing message/auth data.

## Authenticated Windows evidence already established
- WhatsApp / Telegram / LINE authentication: **PASS**;
- GT translation itself is known-good and is the translation oracle for this refactor;
- ordinary WhatsApp private send with outgoing translation **OFF**: **PASS** through the native WhatsApp composer;
- broadcast text: **PASS**;
- broadcast media/file: **PASS**;
- LID/direct addressing: **PASS**;
- group send/mention: **PASS**;
- authenticated reload: **PASS**;
- full exit/restart/session persistence: **PASS**.

The only remaining product acceptance item is ordinary WhatsApp private send with outgoing translation **ON** on the current exact head.

## Live failures that forced the architecture change
Earlier generations exposed two independent problems in the real client:

1. a legacy translated-send owner could remain on the native composer path and emit the old generic `翻译失败，原文未发送` notice even when the newer controller was injected;
2. after the generation-4 authority cutover, a real private-chat test with outgoing translation enabled sent the **raw source text without translating it**. Translation OFF still sent normally.

That second result proved the remaining design was still too complex and, more importantly, violated the fail-closed contract. The likely boundary was chat-scoped configuration identity: the settings UI can key an override by the active WA-JS LID while the native send seam can arrive with the same person represented as PN/c.us. Treating those IDs as unrelated made the adapter incorrectly classify translation as OFF and pass raw text through.

## Thin shared-translation adapter (#665)
The current implementation deliberately follows the same product model as GT/TG/LINE:

`platform send -> resolve platform chat identity/config -> shared translation request -> platform native send`

WhatsApp-specific behavior is now limited to platform adaptation:

- `ui/whatsapp-direct-composer-controller.js` is generation **6**;
- it has **no DOM Enter/click gesture owner** and does not use a second public `WPP.chat.sendTextMessage` transport for translated private sends;
- the existing low-level native WhatsApp send seam delegates to `handleNativeSend()`;
- when outgoing translation does not apply, the captured native send is passed through unchanged;
- when outgoing translation applies, the adapter calls the existing shared `__geekTranslationRequest`, replaces only the outgoing text argument with the translated text, preserves the remaining native args/options/quoted semantics, then invokes the captured native send exactly once;
- Translation Runtime / Worker / provider / quota behavior is **not reimplemented** in the WhatsApp adapter;
- translation failure, missing config, empty translation, or unresolved identity is fail-closed and never invokes native send with raw source;
- group/non-direct/non-string sends remain outside this private-chat translation adapter;
- legacy WhatsApp translation recovery is not bootstrapped on fresh startup and superseded fallback owners remain inactive.

### LID / PN identity safety
A chat override is reused across LID and PN/c.us only when WA-JS can prove they identify the same contact:

- direct PN IDs normalize to a canonical phone identity;
- LID resolves through established WA-JS `contact.getPnLidEntry()` using `phoneNumber` with legacy `pn` compatibility;
- proven same contact -> active-chat override may be used;
- proven different contact -> that override is never applied to the native send chat;
- unresolved LID/PN relation while a local override could matter -> **fail closed; raw source is not sent**.

The contract explicitly covers the real regression shape: settings stored under an active LID while the native send chat is PN/c.us. It requires one shared translation request, exactly one native send containing the translated text, preserved native options, and zero raw-source leakage.

## Current master sync
Current master includes #664 (`fix: settle saturated translation bridge requests`), which returns a typed `BRIDGE_BUSY` terminal error for authenticated requests rejected by WebView bridge capacity instead of leaving the guest Promise pending. That change only touches `ui/webview-bridge-security.js` and its contract and has been synchronized into this feature candidate.

## Current exact-head validation
Current head: `15c71ae7375d028e50e40e2f465875552cd161ed`

Current base: `master@96b9295da3989b1318b14f824b28f8d496abb4a6`

Hosted validation on this exact combined head:
- repository `test` run `35000382778`: **success**;
- Cloudflare Worker validation run `35000382613`: **success**, including repository tests, deployment-tooling syntax and website/subscription/translation/release Worker dry-runs;
- Electron whitespace: **success**;
- Linux full Electron smoke: **success**;
- Windows deployment closure: **success**;
- Windows WhatsApp cold-start: **success**;
- Windows WA-JS 4.6 runtime gate: **success**, requiring generation **6**, native-send owner, translation-setting resolver, LID/PN identity verifier, **no DOM gesture owner**, recovery absence and legacy fallback inactivity;
- aggregate `electron-e2e` run `35000382669`: **success**;
- `validation-client-build` run `35000382668`: **skipped as expected** because this is not a formal client release.

Child PR #665 also completed exact-head validation. Its first Linux smoke attempt hit the independent lock-screen focus-settle flake; one controlled rerun on the **same source SHA** passed the full Linux smoke, while Windows generation-6 runtime and the other gates were already green. The final combined #573 head above passed Linux smoke directly.

## Final remaining live boundary
Keep this PR Draft. On Windows, run Geek at exact head `15c71ae7375d028e50e40e2f465875552cd161ed`, fully exit/restart Geek first, keep WhatsApp authenticated, enable **发送翻译**, then send one ordinary private-chat test message to a controlled recipient.

Acceptance:
1. recipient receives exactly one correct translated message;
2. raw source is not sent separately;
3. correct recipient receives it;
4. no duplicate send occurs.

If the adapter cannot prove LID/PN identity or the shared translation request fails, it must block the send rather than release the raw source.

Do not mark Ready or merge #573 to master until this translated private-chat live acceptance succeeds.

## Scope boundary
No formal client release has been triggered. `.github/release-client-version` is unchanged. No Worker production deployment, R2 upload, D1 migration, data migration, version bump or production mutation is authorized or performed here.

Refs #522, #516, #617, #618, #633, #635, #636, #640, #645, #650, #653, #659, #662, #663, #664, #665, #667.


9529360-cpu:
Maintenance review on current live repo found two follow-ups before this compatibility migration should be considered merge-ready:

1. The migration contract hard-coded client version `1.2.25`; that has now been corrected on this branch to assert package/release-marker synchronization instead, so the test will survive the next legitimate release.
2. `injectWppWithRetry()` currently requires the entire WAPLUS compatibility surface before marking an already-ready official WA-JS 4.6 runtime as injected. The rest of this PR deliberately switches active callers to `window.WPP || window.WAPLUS_WPP`, and current active callers do not require WAPLUS to be complete when official WPP is healthy. This over-gating should be resolved or explicitly justified with an executable fallback-dependency oracle; otherwise a compatibility-layer miss can unnecessarily disable the official primary path.

Separately, the PR's own evidence boundary remains valid: synthetic CI does not substitute for authenticated ordinary send, broadcast text/media, LID/direct chat, group mention, reload and restart/session-persistence evidence. Keeping this PR non-mergeable by maintenance policy until those boundaries are closed.

9529360-cpu:
Maintenance review update (no code mutation from this reviewer):

1. The current branch still encodes the known over-gating blocker in **two places**, not just `injectWppWithRetry()`: `runtimeReady()` in `whatsapp-wa-js-runtime.e2e.cjs` requires `fallbackReady === true`, so the synthetic gate cannot prove the intended architecture where official WA-JS is primary and WAPLUS is a compatibility fallback. If the primary surface is complete but WAPLUS fails, both product injection and CI currently reject the session.

2. Upstream risk changed after this branch was opened: WA-JS 4.6.0 is still the latest stable and WPPConnect 2.3.0 uses it, but upstream currently has open regression #3618 (`isAuthenticated/ChatStore/Conn/MsgStore not found` on WhatsApp Web `2.3000.1046379654`, reproduced on 4.6.0/nightly/wppconnect) and older related #3495. Therefore the migration must not treat the version pin itself as proof that `ChatStore`/module readiness is stable. The existing runtime readiness probe is useful, but its failure/degraded semantics need to remain explicit.

Recommended contract before Ready:
- `primaryReady`: exact WA-JS version + Geek-consumed official surface (`loader/module map`, text/media/active-chat, LID/group, stores).
- `fallbackReady`: separate compatibility evidence only for call sites that truly require WAPLUS-specific behavior.
- `injectWppWithRetry()` should mark the account usable when `primaryReady` is true; a missing WAPLUS layer should be logged/diagnosed and only disable the specific compatibility-only capability, unless repository evidence proves an essential Geek path still has no primary equivalent.
- E2E should have a positive control for `primaryReady=true, fallbackReady=false` and prove primary-backed ordinary/broadcast text paths remain admitted; keep a separate fallback compatibility assertion for the paths that need it.
- Keep authenticated Windows evidence as the final boundary, especially because upstream 4.6.0 still has current module-store regressions on newer WA Web builds.

No suggestion to merge yet; branch also remains stale against current master and shares `ui/app.js` with #524/#518.

9529360-cpu:
Additional lifecycle finding from upstream + current branch wiring:

`injectWppWithRetry()` is invoked from the WhatsApp guest's `did-finish-load`, including the pre-auth/login/QR lifecycle. Upstream WA-JS's documented bootstrap separates these states: it waits for `window.WPP?.isReady`, then separately calls `WPP.auth.isAuthenticated()`. In other words, WA-JS injection readiness is not the same contract as authenticated business-capability readiness.

This branch currently makes the *initial injection owner* wait for the full surface (`loader/module map`, `sendTextMessage`, `sendFileMessage`, `getActiveChat`, LID/group, `ChatStore`, `UserPrefs`) before adding the partition to `wppInjected`, and retries/reinjects the entire bundle otherwise. That conflates:

1. WA-JS injected/module graph settled;
2. authenticated WhatsApp runtime capabilities available;
3. WAPLUS compatibility surface available.

That matters more now because upstream 4.5.0 explicitly fixed cold-load module-graph settling, while current open upstream regressions still report `ChatStore`/other modules missing on newer WA Web even with 4.6.0. A real login/QR page or transient module drift could therefore trigger repeated whole-bundle injection although WA-JS itself is already successfully installed.

Recommended state split before Ready:
- **injectionReady**: `WPP.isReady` (+ only loader facts truly required to prove the bundle settled). This owns `wppInjected` / duplicate-injection prevention.
- **primaryCapabilityReady**: the exact official APIs Geek needs for a given operation, checked at/near the operation or by a separate capability probe; failure is a classified/degraded capability, not automatic whole-bundle reinjection.
- **fallbackCapabilityReady**: WAPLUS-specific compatibility evidence, independent from official primary admission.
- keep authenticated Windows cold-start/reload/send evidence as final acceptance.

This also gives a stronger E2E oracle: prove one injection is retained while capability readiness can be absent/recover later, instead of requiring every business store to exist before the injection lifecycle is considered complete.

9529360-cpu:
Maintenance review finding — WA-JS readiness owner must be split before this Draft becomes merge-ready.

Current `injectWppWithRetry()` still conflates four different states: (1) bundle/bootstrap installed (`WPP.isInjected` + loader connectivity), (2) main WA modules ready (`WPP.isReady`), (3) authenticated/session-dependent business capability readiness, and (4) optional WAPLUS fallback capability readiness.

Upstream WA-JS intentionally separates `isInjected`, `isReady` and later module/auth readiness. Its `exportModule` layer is lazy: first access searches the WhatsApp module graph and then caches the discovered value (or a not-found/undefined result). That matters here because probing the WAPLUS function surface while the main module graph is not ready can freeze an early negative lookup, while the current retry loop responds by re-evaluating the entire WA-JS bundle.

Required migration shape:

1. Treat `WPP.isInjected === true` plus loader/module-map connectivity as bootstrap ownership. Once true, do **not** reinject the WA-JS bundle merely because login/business modules are not ready yet.
2. Before `WPP.isReady`, do not probe WAPLUS lazy function getters.
3. After `WPP.isReady`, validate the exact Geek-consumed official `window.WPP` primary surface. If that official surface is healthy, accept it without requiring the full WAPLUS fallback surface.
4. Initialize/probe WAPLUS only when the official primary is missing a capability that an existing compatibility path actually needs; fallback readiness is a separate state, not a condition for declaring a healthy official primary unusable.
5. Keep auth/session readiness distinct from duplicate-injection prevention. A pre-auth QR page must be allowed to stay `isInjected=true,isReady=false` without triggering whole-bundle reinjection.
6. Add executable coverage for: pre-auth injected/not-ready without reinjection; official-primary-ready + fallback-incomplete accepted; fallback lazy getters not touched pre-ready; capability loss/recovery without duplicate bundle injection.

This also means the current synthetic runtime oracle requiring `fallbackReady === true` for overall runtime readiness is too strict and should be split into primary-ready vs compatibility-fallback evidence.

No recommendation to merge yet: branch still needs deliberate refresh onto live master and authenticated Windows WhatsApp evidence for ordinary send, broadcast text/media, LID/direct chat, group mention paths, reload and restart/session persistence.

9529360-cpu:
2026-09-15 readiness review — upstream/local compatibility evidence

The two remaining semantic blockers can now be stated more precisely:

1. Upstream WA-JS itself defines three different loader stages in `src/loader/index.ts`: `isInjected` (loader injected), `isReady` (main loader modules loaded), and `isFullReady` (all runtime modules loaded). Geek should not collapse those loader stages into authenticated/business-capability readiness.
2. WA-JS 4.5 explicitly fixed stale negative module lookup on modern Meta-loader cold starts (#3419/#3499). Current 4.6 source keeps a module-count-aware negative-cache invalidation path: a cached null is re-scanned when new modules register. Therefore do **not** describe official WA-JS 4.6 as permanently poisoning an early module miss.
3. The checked-in `resources/waplus-wpp.js` compatibility bundle is a separate concern. Its bundled module-search code visibly caches a miss with `cache.set(condition, null)`, and repository inspection did not find the upstream `searchIdMissModuleCount` invalidation marker. Treat this local compatibility bundle as an independently versioned/older discovery layer until proven otherwise.

Implementation consequence:
- injection ownership should settle on official WA-JS loader state, not on WAPLUS business surface;
- official WPP primary capabilities should be evaluated after the appropriate WA-JS readiness stage and authentication context;
- WAPLUS should be initialized/probed only as a compatibility fallback for a demonstrated missing primary capability, not as a prerequisite for declaring the official runtime injected;
- the synthetic runtime gate should report these states separately (`injected`, `mainReady`, `fullReady`, primary capability state, fallback capability state) rather than one `runtimeReady` boolean.

This strengthens the existing Draft blockers; it does not authorize merge. Authenticated Windows ordinary send + broadcast/media/LID/group/reload/restart evidence remains required.

9529360-cpu:
Superseded by fresh-current-master candidate #700.

#700 was rebuilt from live `master@2805d0f499618b4aa41b4ee8b0e2aef9a7c0db03` instead of carrying this branch's stale 99-commit history. Current #700 exact head is `41923e4ef374087f0037aecd517ff041969ddfac`.

The refresh preserved the audited WA-JS 4.6 migration and also found/fixed one integration regression that did not exist when this branch was originally cut: generation-6 private translated sends omitted explicit Translation Runtime intent and would therefore be classified as background `message-display` work by current master. #700 now sends `intent: 'outgoing-send'`, bumps the guest owner to generation 7, and locks that contract in both focused and send-safety tests.

Hosted exact-head evidence on #700 is green: repository test, Cloudflare Worker validation/dry-runs, whitespace, Windows deployment closure + WhatsApp cold-start + WA-JS 4.6 generation-7 runtime gate, Linux Electron smoke, and aggregate Electron E2E. The validation client build remains skipped because the formal client version/marker remains 1.2.27.

Historical authenticated evidence recorded on this PR remains useful archaeology, but it is not evidence for the new exact head. The remaining product acceptance boundary has moved to #700: on an authenticated Windows client at that exact candidate, after full exit/restart and with outgoing translation enabled, one ordinary private-chat send must arrive exactly once as the correct translation to the correct recipient, with no raw-source duplicate. Keep #700 Draft until that real-client check succeeds; synthetic CI does not substitute for it.

Issue #444, opened 2026-09-14
ops: enforce master branch protection and required checks


9529360-cpu:
## Problem

Repository maintenance already follows an exact-head PR/CI discipline in practice, but live branch metadata still shows the `master` protection enforcement/status-check enforcement as off. That leaves an important repository invariant dependent on maintainer behavior instead of repository rules.

This is an owner/admin control-plane task, not a product-code change.

## Goal

Make `master` enforce the merge discipline that the repository already expects, without creating path-filter deadlocks or weakening current CI.

## Acceptance criteria

- direct unreviewed pushes to `master` are not the normal path;
- force-push and branch deletion are blocked unless an explicitly documented emergency/admin bypass applies;
- PR merges require the always-emitted core gates that protect desktop changes, at minimum the repository `test` gate and aggregate `electron-e2e` gate;
- any path-filtered check such as `cloudflare-worker-validation` is not made required until its trigger/reporting model guarantees a terminal check for every PR that would otherwise need it; do not create a permanent pending-check deadlock;
- required-check names are unique and match the actual GitHub Actions check names;
- strict/up-to-date behavior and admin/bypass policy are explicitly reviewed rather than assumed;
- verify the final rule with a harmless PR: a missing/failing required gate must block merge and a fully green PR must remain mergeable.

## Safety boundary

Do not disable CI, weaken Electron/WebView safety checks, expose repository secrets, or change client release behavior as part of this task. This requires repository administration/owner permissions; routine code-maintenance credentials must not be expanded just to bypass the control plane.

Reference: GitHub protected-branch/required-status-check rules should be configured in repository rules rather than remembered as prose by each maintainer.

9529360-cpu:
2026-09-14 live-state reconciliation: this control-plane task is already implemented through the repository ruleset layer, so the Issue's old statement that `master` enforcement is off is stale.

Live repository ruleset `Protect master` (id `22623013`) is **active** and targets exactly `refs/heads/master`. It currently enforces:

- branch deletion blocked;
- non-fast-forward / force-push blocked;
- pull request required before merging;
- review-thread resolution required;
- required GitHub Actions checks `test` and `electron-e2e` (integration id `15368`);
- strict required-status-check policy (`master` freshness required);
- no bypass actors; current user cannot bypass.

The required-check workflows are also safe to require:

- `.github/workflows/test.yml` runs on every pull request without PR `paths` / `paths-ignore` filtering and exposes the required job name `test`;
[...]

PR #575, opened 2026-09-14
docs: make governance references self-validating

9529360-cpu:
Closes #574.

## Verified governance gap

Durable/current maintainer Markdown had no deterministic repository-local reference contract, so a rename/prune could silently leave a current runbook pointing at a missing `.md` file. Separately, `docs/github-control-plane.md` still described master protection as pending owner/admin work tracked by #444 even though #444 is completed and the live repository has an active `Protect master` ruleset.

The original audit's inference that several docs were missing was caused by a truncated recursive API response and is explicitly withdrawn in #574; this PR addresses only the verified gaps above.

## Final change

- add one deterministic, network-free `test/documentation-authority-contract.cjs` to the normal repository suite;
- scan the explicit durable/current governance + runbook set, not historical research/prototype documents;
- validate repository-local Markdown link destinations relative to the source file;
- validate backticked `.md` paths using repository-root semantics first, then source-relative fallback;
- reject references that escape the repository or point to missing files;
- ignore external/scheme URLs, pure anchors and fenced code examples;
- change only the merge-control section of `docs/github-control-plane.md`: #444 becomes completed rollout/history, while future maintainers must query live repository rulesets for exact enforcement/check parameters.

## Mature precedent

This uses the same boundary as `remark-validate-links` and VS Code's Markdown validator: local file references can be checked deterministically/offline, while external URL liveness is a separate concern. Geek keeps this narrow check dependency-free rather than adding a general Markdown toolchain.

## Freshness / scope

Refreshed onto live `master` `d166aee47ee650752154675fc0f01657ce7c61cc`; compare is `behind=0`. Final diff is exactly two files: 4 lines in the control-plane runbook plus the focused contract. No Runtime, Worker, Subscription, WebView, broadcast, workflow, package version or release-marker changes.

Exact-head `test` and aggregate `electron-e2e` are required before merge.

PR #579, opened 2026-09-14
test: make durable documentation references self-validating

### LINKED ISSUE #574: [P2][Architecture] Make durable documentation references self-validating
## Corrected finding

The initial audit incorrectly inferred that several `docs/` files were missing because a recursive GitHub API response was truncated. Direct file lookup/code search confirmed the referenced runbooks (including `docs/windows-real-client-runner.md` and `docs/群发最终实现约束-20260824.md`) are present. That false premise is withdrawn here explicitly.

The real remaining governance gap is narrower:

1. durable repository documentation has no deterministic contract that prevents future repository-local Markdown references from silently becoming broken after a rename/prune;
2. `docs/github-control-plane.md` still describes `master` protection / required checks as owner/admin follow-up tracked by #444, while #444 is completed and the live `Protect master` ruleset is active. The exact ruleset remains dynamic and must still be queried live.

## Target

- Add a deterministic, network-free contract that scans the durable governance Markdown set and fails when a repository-local `.md` link/backticked path points outside the repo or to a missing file.
- Ignore external URLs, anchors and fenced code examples.
- Correct the stale merge-control prose so #444 is treated as completed rollout/history, while future maintainers are still required to query the live repository ruleset before relying on its exact parameters.
- Do not create another status database or hard-code changing HEAD/version/PR/CI state.
- No production runtime, Worker, package version or release-marker changes.

## Validation

- focused documentation-authority contract;
- full repository `test` gate;
- aggregate `electron-e2e` gate required by the live `master` ruleset.

## Collision boundary

Do not touch Translation Runtime/Worker, Subscription, gateway routing, WebView runtime, broadcast runtime or active translation PR write sets.

9529360-cpu:
Closes #574.

## Root cause

Durable maintainer/runbook Markdown had no deterministic repository-local reference contract. A rename or prune could silently leave current governance docs pointing at a missing `.md` file until a human noticed. Separately, `docs/github-control-plane.md` still described master protection as pending owner/admin work tracked by #444 even though #444 is completed and the live repository has an active `Protect master` ruleset.

## Change

- add `test/documentation-reference-authority-contract.cjs` to the normal repository test suite;
- scan only the explicit durable/current governance set rather than historical research/prototype docs;
- validate local Markdown link destinations relative to the source file;
- validate backticked `.md` paths using repository-root semantics first, then source-relative fallback;
- fail closed when a local reference escapes the repository or points to a missing file;
- ignore external/scheme URLs, pure anchors and fenced code examples;
- update merge-control prose so #444 is completed rollout/history, while exact live ruleset parameters must still be queried from GitHub.

The contract intentionally does not check the network or turn Markdown into another live status database.

## Mature precedent

The boundary follows `remark-validate-links` and VS Code Markdown link validation: repository-local file/fragment validation can be deterministic and offline, while external URL liveness is a separate concern. Geek keeps the implementation dependency-free because the required scope is deliberately narrower than a general Markdown processor.

## Scope

Exactly two files. No Runtime, Worker, Subscription, WebView, broadcast, package version, release marker, workflow or production deployment input changes. Exact-head repository test and the repository's Electron aggregate gate are required before merge.


9529360-cpu:
Closing as duplicate of existing PR #575, which already owns Issue #574 and the same documentation-reference write set. The later generic open-PR refresh surfaced #575 after an earlier narrow keyword search missed it. Per the repository's single-owner rule, maintenance continues on #575 only; no parallel implementation will be kept.

